PR #22, opened 2022-05-24
Fix deprecation condition

kklein:
The method `is_deprecated` no longer exists for class `Condition`.


ivergara (approved):
I like when some cleanup is done!

PR #23, opened 2022-05-27
Add: Kolmogorov-Smirnov Test

YasinTatarQC:
Hey!

In order to catch distributional shifts in our data-driven projects, I'd like to introduce a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) to datajudge.

Since this introduces a new `stats.py` subpackage (and this is my first time contributing to datajudge) I'd love to hear your feedback on the following possible issues I see (hence the _draft_ status):

- [x] Afaik, this is the first test that needs the full data column from the database. Therefore, I have introduced `_column(...)` and `get_data(...)` to `db_access`. Is there a better way to do this? `_column(...)` shares a lot with `_column_aggreate(...)` of course.
- [x] Introducing `scipy` as a new dependency just for this test is a bit much. With colleagues, I have discussed options like 
    - (1) just implementing the KS test myself again 
    - (2) stating `scipy` as an optional dependency and warning the user. Any other ideas?
- [x] More tests are needed, in my opinion. Am working on that!

Thanks!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Changelog
 
 **New feature: Statistical Tests**
 
-- Implemented a new constraint :class:`~datajudge.constraints.stats.KolmogorovSmirnov2Sample` for :class:`~datajudge.BetweenRequirement` that performs a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) between two data sources.
+- Implemented a new constraint :class:`~datajudge.constraints.stats.KolmogorovSmirnov2Sample` for :class:`~datajudge.BetweenRequirement` that performs a `Kolmogorov Smirnov Test <https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test>`_ between two data sources.
 
 1.0.1 - 2022.05.24
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,10 +14,6 @@ Changelog
 
 - Implemented a new constraint :class:`~datajudge.constraints.stats.KolmogorovSmirnov2Sample` for :class:`~datajudge.BetweenRequirement` that performs a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) between two data sources.
 
-**New method: get_column**
-
-- :meth:`db_access.get_column` allows retrieving the full data column, i.e. without applying any operations.
-
 1.0.1 - 2022.05.24
 ------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ Changelog
 
 **New feature: Statistical Tests**
 
-- Implemented a :class:`~datajudge.BetweenRequirement` constraint running a Kolmogorov Smirnov test.
+- Implemented a new constraint :class:`~datajudge.constraints.stats.KolmogorovSmirnov2Sample` for :class:`~datajudge.BetweenRequirement` that performs a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) between two data sources.
 
 **New method: get_column**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,17 @@
 Changelog
 =========
 
+1.1.0 - 2022.06.01
+------------------
+
+**New feature: Statistical Tests**
+
+- Implemented a :class:`~datajudge.BetweenRequirement` constraint running a Kolmogorov Smirnov test.
+
+**New method: get_column**
+
+- :meth:`db_access.get_column` allows retrieving the full data column, i.e. without applying any operations.
+
 1.0.1 - 2022.05.24
 ------------------
 
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -12,7 +12,7 @@ class KolmogorovSmirnov2Sample(Constraint):
     This is detected by applying the two-sample Kolmogorov-Smirnov test.
     """
 
-    def __init__(self, ref: DataReference, ref2: DataReference, significance_level: float = 0.01):
+    def __init__(self, ref: DataReference, ref2: DataReference, significance_level: float = 0.05):
         self.significance_level = significance_level
         super().__init__(ref, ref2=ref2)
 
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -11,10 +11,6 @@ from .. import db_access
 
 
 class KolmogorovSmirnov2Sample(Constraint):
-    """
-    This constraint assures that two given given data references are sampled from the same underlying distribution.
-    This is detected by applying the two-sample Kolmogorov-Smirnov test.
-    """
 
     def __init__(
         self, ref: DataReference, ref2: DataReference, significance_level: float = 0.05
@@ -49,7 +45,7 @@ class KolmogorovSmirnov2Sample(Constraint):
         result = p_value >= self.significance_level
         assertion_text = (
             f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
-            f"has p-value {p_value}"
+            f"has p-value {p_value}  < {self.significance_level}"
             f"{self.condition_string}"
         )
 
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -11,7 +11,6 @@ from .. import db_access
 
 
 class KolmogorovSmirnov2Sample(Constraint):
-
     def __init__(
         self, ref: DataReference, ref2: DataReference, significance_level: float = 0.05
     ):
@@ -29,10 +28,6 @@ class KolmogorovSmirnov2Sample(Constraint):
 
         # calculate statistic
         statistic, p_value = ks_2samp(data, data2)
-
-        sql_query = """
-            < BIG SQL >
-        """
 
         return p_value
 
@@ -54,11 +49,3 @@ class KolmogorovSmirnov2Sample(Constraint):
         )
 
         return result, assertion_text
-
-    # test() -> get_factual + get_target
-    # get_factual -> retrieve(ref)
-    # get_target -> retrieve(ref2)
-    # test: compare(factual, target)
-
-    # get_factual -> runs sql query -> retrieves p_value
-    # get_target -> returns the TARGET p_value
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -13,7 +13,7 @@ class KolmogorovSmirnov2Sample(Constraint):
     """
 
     def __init__(self, ref: DataReference, ref2: DataReference, significance_level: float = 0.01):
-        self.significance_leve = significance_level
+        self.significance_level = significance_level
         super().__init__(ref, ref2=ref2)
 
     @staticmethod
@@ -43,14 +43,14 @@ class KolmogorovSmirnov2Sample(Constraint):
         return p_value
 
     def get_target_value(self, engine: sa.engine.Engine) -> Any:
-        return self.significance_leve
+        return self.significance_level
 
     def compare(
             self, value_factual: Any, value_target: Any
     ) -> Tuple[bool, Optional[str]]:
         # value_factual := calculated p-value from data
         # value_target := required significance level provided by user
-        result = value_factual > value_target
+        result = value_factual >= value_target
         assertion_text = (
             f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
             f"has p-value {value_factual}"
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,9 +1,10 @@
 from typing import Any, Tuple, Optional, Callable
-
+from .. import db_access
 from datajudge import Constraint
+from datajudge.constraints.base import OptionalSelections
 from datajudge.db_access import DataReference
 import sqlalchemy as sa
-from scipy.stats import ks_2samp, kstest
+from scipy.stats import ks_2samp
 
 
 class KolmogorovSmirnov2Sample(Constraint):
@@ -30,30 +31,20 @@ class KolmogorovSmirnov2Sample(Constraint):
 
         return p_value
 
-    def get_factual_value(self, engine: sa.engine.Engine) -> Any:
-        # get data
-        data = self.retrieve(engine, self.ref)
-        data2 = self.retrieve(engine, self.ref2)
-
-        # clean data
-
-        # calculate test statistics
-        p_value = self.calculate_2sample_ks_test(data, data2)
-
-        return p_value
-
-    def get_target_value(self, engine: sa.engine.Engine) -> Any:
-        return self.significance_level
+    def retrieve(
+        self, engine: sa.engine.Engine, ref: DataReference
+    ) -> Tuple[Any, OptionalSelections]:
+        return db_access.get_data(engine, ref)
 
     def compare(
             self, value_factual: Any, value_target: Any
     ) -> Tuple[bool, Optional[str]]:
-        # value_factual := calculated p-value from data
-        # value_target := required significance level provided by user
-        result = value_factual >= value_target
+        # factual and target values are the corresponding columns from our data source
+        p_value = self.calculate_2sample_ks_test(value_factual, value_target)
+        result = p_value >= self.significance_level
         assertion_text = (
             f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
-            f"has p-value {value_factual}"
+            f"has p-value {p_value}"
             f"{self.condition_string}"
         )
 
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -3,11 +3,9 @@ from typing import Any, Optional, Tuple
 import sqlalchemy as sa
 from scipy.stats import ks_2samp
 
-from datajudge import Constraint
-from datajudge.constraints.base import OptionalSelections
-from datajudge.db_access import DataReference
-
 from .. import db_access
+from ..db_access import DataReference
+from .base import Constraint, OptionalSelections
 
 
 class KolmogorovSmirnov2Sample(Constraint):
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -48,7 +48,9 @@ class KolmogorovSmirnov2Sample(Constraint):
     def compare(
             self, value_factual: Any, value_target: Any
     ) -> Tuple[bool, Optional[str]]:
-        result = value_factual <= value_target
+        # value_factual := calculated p-value from data
+        # value_target := required significance level provided by user
+        result = value_factual > value_target
         assertion_text = (
             f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
             f"has p-value {value_factual}"
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -30,6 +30,10 @@ class KolmogorovSmirnov2Sample(Constraint):
         # calculate statistic
         statistic, p_value = ks_2samp(data, data2)
 
+        sql_query = """
+            < BIG SQL >
+        """
+
         return p_value
 
     def retrieve(
@@ -50,3 +54,11 @@ class KolmogorovSmirnov2Sample(Constraint):
         )
 
         return result, assertion_text
+
+    # test() -> get_factual + get_target
+    # get_factual -> retrieve(ref)
+    # get_target -> retrieve(ref2)
+    # test: compare(factual, target)
+
+    # get_factual -> runs sql query -> retrieves p_value
+    # get_target -> returns the TARGET p_value
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,7 +1,6 @@
 from typing import Any, Optional, Tuple
 
 import sqlalchemy as sa
-from scipy.stats import ks_2samp
 
 from .. import db_access
 from ..db_access import DataReference
@@ -21,6 +20,13 @@ class KolmogorovSmirnov2Sample(Constraint):
         For two given lists of values calculates the Kolmogorov-Smirnov test.
         Read more here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html
         """
+        try:
+            from scipy.stats import ks_2samp
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError(
+                "Calculating the Kolmogorov-Smirnov test relies on scipy."
+                "Therefore, please install scipy before using this test."
+            )
 
         # calculate statistic
         statistic, p_value = ks_2samp(data, data2)
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -35,7 +35,7 @@ class KolmogorovSmirnov2Sample(Constraint):
     def retrieve(
         self, engine: sa.engine.Engine, ref: DataReference
     ) -> Tuple[Any, OptionalSelections]:
-        return db_access.get_data(engine, ref)
+        return db_access.get_column(engine, ref)
 
     def compare(
         self, value_factual: Any, value_target: Any
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -21,8 +21,6 @@ class KolmogorovSmirnov2Sample(Constraint):
         For two given lists of values calculates the Kolmogorov-Smirnov test.
         Read more here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html
         """
-        # data validation:
-        # TODO: perform checks that are needed for scipy's test function
 
         # calculate statistic
         statistic, p_value = ks_2samp(data, data2)
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -31,7 +31,11 @@ class KolmogorovSmirnov2Sample(Constraint):
                 "Therefore, please install scipy before using this test."
             )
 
-        # calculate statistic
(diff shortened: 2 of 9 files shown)


ivergara:
> Hey!

Hey!

> In order to catch distributional shifts in our data-driven projects, I'd like to introduce a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) to datajudge.

Thank you for your contribution! I do think it's worth having such constraints being implemented.
 
> Since this introduces a new `stats.py` subpackage (and this is my first time contributing to datajudge) I'd love to hear your feedback on the following possible issues I see (hence the _draft_ status):
> 
> * [ ]  Afaik, this is the first test that needs the full data column from the database. Therefore, I have introduced `_column(...)` and `get_data(...)` to `db_access`. Is there a better way to do this? `_column(...)` shares a lot with `_column_aggreate(...)` of course.

I don't see an issue with having `_column` and `_column_aggregate`, they could be merged and have a keyword flag to control aggregation. But, I don't see it as something really critical. However, I don't see much use for the `get_data` function.

> * [ ]  Introducing `scipy` as a new dependency just for this test is a bit much. With colleagues, I have discussed options like
>   
>   * (1) just implementing the KS test myself again
>   * (2) stating `scipy` as an optional dependency and warning the user. Any other ideas?

Yes, it's a bit much. However, we already have as a "dependency" `snowflake-sqlalchemy` which is rather big too and we can follow the same approach here.

On the other hand, as you state, this constraint implies doing a full copy of the data of two columns locally. I find that somewhat of an issue if we want to handle tables with hundred of millions of rows. 

Now, if you tell me to implement KS yourself, I'd say it'd be a waste to do it in Python so better see if there is a way to implement it at least partially within the confines of the database engine. The challenge here would be to do it in a way that it's database agnostic unless you want to do versions for Postgres, mssql, and snowflake.

Actually, a smarter data access pattern would be highly beneficial even if we keep scipy as the calculator. 

> * [ ]  More tests are needed, in my opinion. Am working on that!

More tests is always good!

YasinTatarQC:
> However, I don't see much use for the `get_data` function.

`get_data` is just there to preserve the current code structure, since you wouldn't want to call _column(...) yourself as a user.
I have now refactored `_column` and removed `_column_aggregate`. Hope that simplifies it a little bit.

YasinTatarQC:
> On the other hand, as you state, this constraint implies doing a full copy of the data of two columns locally. I find that somewhat of an issue if we want to handle tables with hundreds of millions of rows.

Very good point and something I have considered! While Oracle has [KS support built-in ](https://docs.oracle.com/database/121/SQLRF/functions187.htm#SQLRF06319) I couldn't find solutions for MS SQL, Postgres, or Snowflake. There is, however, [this script](https://github.com/spgarbet/sql-kstest) available that would need to be tested.
My solution would be: Check where this script works and create a `ks_supported_databases` list. If our engine is part of this list: nice. If not, warn the user about our scipy dependency and run the test locally on scipy.

src/datajudge/db_access.py
```diff
@@ -632,21 +632,30 @@ def get_row_count(engine, ref, row_limit: int = None):
...
+    if not aggregate_operator:
+        selection = sa.select([column])
+        result = engine.connect().execute(selection).all()
 
-def _column(engine, ref):
-    subquery = ref.get_selection(engine).alias()
-    column = subquery.c[ref.get_column(engine)]
-    selection = sa.select([column])
-    result = engine.connect().execute(selection).all()
+        return [elem[0] for elem in result], [selection]
```
ivergara, line 652: Since you're doing an `else` statement I'd say 

```suggestion
        result =  [elem[0] for elem in result]
```

Together with the next suggestion (couldn't have done it in one block since there is a deleted line in between).

src/datajudge/db_access.py
```diff
@@ -632,21 +632,30 @@ def get_row_count(engine, ref, row_limit: int = None):
...
+
+    else:
+        selection = sa.select([aggregate_operator(column)])
+        result = engine.connect().execute(selection).scalar()
 
-    return [elem[0] for elem in result], [selection]
+        return result, [selection]
```
ivergara, line 658:
```suggestion
    return result, [selection]
```

src/datajudge/db_access.py
```diff
@@ -632,21 +632,30 @@ def get_row_count(engine, ref, row_limit: int = None):
     return result, [selection]
 
 
-def _column_aggregate(engine, ref, column_operator):
+def _column(engine, ref, aggregate_operator: Callable = None):
```
ivergara, line 635: I'm not sure the name of this function reflects very well what's doing.

Also, you could properly annotate all the arguments.

Be mindful on how the docstring style is in the rest of the documentation, I don't remember having seen the style with the `@`'s you're using within the codebase.

ivergara (commented):
Looks good, but this wasn't the biggest issue in your code. Now in this PR besides implementing your new Constraint, you're adding changes in other places that aren't related. That could have been a separate PR refactoring the existing `_column_aggregate` and the new `_column` together.

ivergara:
> > On the other hand, as you state, this constraint implies doing a full copy of the data of two columns locally. I find that somewhat of an issue if we want to handle tables with hundreds of millions of rows.
> 
> Very good point and something I have considered! While Oracle has [KS support built-in ](https://docs.oracle.com/database/121/SQLRF/functions187.htm#SQLRF06319) I couldn't find solutions for MS SQL, Postgres, or Snowflake. There is, however, [this script](https://github.com/spgarbet/sql-kstest) available that would need to be tested. My solution would be: Check where this script works and create a `ks_supported_databases` list. If our engine is part of this list: nice. If not, warn the user about our scipy dependency and run the test locally on scipy.

Using something like that test would be great. From a fast scan, seems it should be supported by any SQL engine.

Concerning the `ks_supported_databases` list/set/container, I wouldn't worry too much. However, it's fine if you want to add that logic and there is already a precedent of some constraints not working for some supported engines (but it's something we should try to minimize and work in eradicating it). We make the statement we support three engines and we test actively on them, so it's somewhat clear that any use of `datajudge` against other databases is at the user's "risk".

src/datajudge/db_access.py
YasinTatarQC: What do you think about `_get_column`? I initially wanted to follow the `_column_aggregate` scheme, but this doesn't translate very well, you're right.

Regarding the Docstring: Do we have a pre-configured style? This was the default in PyCharm.
This [post](https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats) talks about different styles and the PEP 257 discussion. In my opinion, `reStructuredText` looks like the default in the community for me. What is your opinion?

ivergara: > What do you think about `_get_column`? I initially wanted to follow the `_column_aggregate` scheme, but this doesn't translate very well, you're right.

Yes, probably that conveys better the intention of the function.
> 
> Regarding the Docstring: Do we have a pre-configured style? This was the default in PyCharm. This [post](https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats) talks about different styles and the PEP 257 discussion. In my opinion, `reStructuredText` looks like the default in the community for me. What is your opinion?

I believe we tend to use `numpy` version. Also, if you see most of the code, documentation is mostly absent on the constraints themselves and the documentation basically resides in the `add_xxx` methods of the requirements. I'll defer to @kklein for this.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any, Optional, Tuple
+
+import sqlalchemy as sa
+from scipy.stats import ks_2samp
+
+from datajudge import Constraint
```
kklein, line 6:
```suggestion
from .base import Constraint, OptionalSelections
```
Adapting for consistency. The other modules in `/constraints/` also import relative to the current module.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,56 @@
...
+from datajudge.constraints.base import OptionalSelections
+from datajudge.db_access import DataReference
```
kklein, line 8:
```suggestion
from ..db_access import DataReference
```
See above.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any, Optional, Tuple
+
+import sqlalchemy as sa
+from scipy.stats import ks_2samp
+
+from datajudge import Constraint
+from datajudge.constraints.base import OptionalSelections
```
kklein, line 7:
```suggestion
```

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,56 @@
...
+    This is detected by applying the two-sample Kolmogorov-Smirnov test.
+    """
+
+    def __init__(
+        self, ref: DataReference, ref2: DataReference, significance_level: float = 0.05
+    ):
+        self.significance_level = significance_level
+        super().__init__(ref, ref2=ref2)
+
+    @staticmethod
```
kklein, line 25: Nice. :)

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def compare(
+        self, value_factual: Any, value_target: Any
+    ) -> Tuple[bool, Optional[str]]:
+        # factual and target values are the corresponding columns from our data source
+        p_value = self.calculate_2sample_ks_test(value_factual, value_target)
+        result = p_value >= self.significance_level
+        assertion_text = (
+            f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
+            f"has p-value {p_value}"
```
kklein, line 52:
```suggestion
            f"has p-value {p_value} < {self.significance_level}"
```
The assertion text assumes failure.

src/datajudge/db_access.py
```diff
@@ -632,43 +632,65 @@ def get_row_count(engine, ref, row_limit: int = None):
...
+    @return: Scalar value if aggregate_operator is requested, otherwise list of values in column
+    """
     subquery = ref.get_selection(engine).alias()
     column = subquery.c[ref.get_column(engine)]
-    selection = sa.select([column_operator(column)])
-    result = engine.connect().execute(selection).scalar()
+
+    if not aggregate_operator:
+        selection = sa.select([column])
+        result = engine.connect().execute(selection).all()
```
kklein, line 650: Maybe it might make sense to use `.scalars()`instead of `.all()` here. :)
https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=fetchall#sqlalchemy.engine.Result.scalars

src/datajudge/requirements.py
```diff
@@ -1251,3 +1252,23 @@ def add_row_matching_equality_constraint(
...
+        self,
+        column1: str,
+        column2: str,
+        condition1: Condition = None,
+        condition2: Condition = None,
+        significance_level: float = 0.05,
+    ):
+        """
+        Apply the so-called two-sample Kolmogorov-Smirnov test to compare the distributions of the given DataSources.
+        This contraint is fulfilled, when the resulting p-value of the Kolmogorov-Smirnov test is higher than the
```
kklein, line 1266: I think something got lost here.

tests/integration/test_integration.py
```diff
@@ -1755,3 +1755,27 @@ def test_groupby_aggregation_within_with_failures(
...
+        (identity, "col_int", "col_int", None, 1.0),
+        (identity, "col_int", "col_int", Condition("col_int >= 3"), 1.0),
+    ],
+)
+def test_ks_2sample_constraint_perfect_between(engine, int_table1, int_table2, data):
+    """
+    Test Kolmogorov-Smirnov for the same column -> p-value should be perfect 1.0.
+    """
+    (operation, col_1, col_2, condition, significance_level) = data
+    req = requirements.BetweenRequirement.from_tables(*int_table1, *int_table1)
```
kklein, line 1772: Should the second argument be `int_table2`? If not, we could remove it from the function parameters.

src/datajudge/db_access.py
```diff
@@ -632,43 +632,65 @@ def get_row_count(engine, ref, row_limit: int = None):
...
+        result = [elem[0] for elem in result]
+
+    else:
+        selection = sa.select([aggregate_operator(column)])
+        result = engine.connect().execute(selection).scalar()
+
     return result, [selection]
 
 
+def get_data(engine, ref):
```
kklein, line 661: Could this potentially be called `get_column` - or similar - instead?

src/datajudge/db_access.py
kklein: > Also, if you see most of the code, documentation is mostly absent on the constraints themselves and the documentation basically resides in the add_xxx methods of the requirements

Yes. The heuristic I've been trying to follow most recently was: 
- Enduser information goes to the `Requirement`'s `add_*` method - this often revolves around what the constraint is about and how to use it.
- Developer information goes into the `Constraint` definition - this often revolves around how the constraint _works_.

kklein:
Hi @YasinTatarQC - thanks a lot for this draft! I'd be very excited to have a ks test be part of datajudge.

I agree with a lot of the things which have been said. Here a few points from my side:
- I think it is not strictly necessary to have the test be done in database, even if this restricts what kind of data the constraint can be used on.
- Yet, I think it would be very cool to have the test be done in database and merely return the 'result' of the test.
- If the test is implemented in database, it would be cool to have a generic, sqlalchemy solution. If this is not in reach, merely getting it to work for postgres, mssql and snowflake is perfectly acceptable from my point of view.

On a different note, I'm currently not sure why the documentation build is failing.

tests/integration/test_integration.py
YasinTatarQC: I wanted to test the same data twice to get a perfect p_value of 1.0, and therefore I need the same tables.
But you're right; the second argument is not needed!

tests/integration/test_integration.py
```diff
@@ -1755,3 +1755,50 @@ def test_groupby_aggregation_within_with_failures(
...
+        significance_level=significance_level,
+    )
+
+    assert operation(req[0].test(engine).outcome)
+
+
+@pytest.mark.parametrize(
+    "data",
+    [
+        (identity, "col_int", "col_int"),
```
kklein, line 1787: Maybe we can add a parameter set here where the result 'swaps', i.e. on the threshold between sufficient and insufficient statistic value.

ivergara:
> Hi @YasinTatarQC - thanks a lot for this draft! I'd be very excited to have a ks test be part of datajudge.
> 
> I agree with a lot of the things which have been said. Here a few points from my side:
> 
> * I think it is not strictly necessary to have the test be done in database, even if this restricts what kind of data the constraint can be used on.

Upon further discussion, we agreed on continuing doing the "stupid" thing and pulling all the data and doing the computation on Python world. Lets not fall into premature optimization yet, there is precedence of doing the stupid thing in other constraints.

> * Yet, I think it would be very cool to have the test be done in database and merely return the 'result' of the test.

If the decision results in some inconvenient runtime at some point, we can try to think on how to do it smarter. One option is to see if we can do an aggregation step in the database, basically pre-generating the CDF, and then compute locally. That'd reduce massively the data transfer. As a last resort, we have moving the full computation to the database using something akin to the SQL script mentioned earlier. 

However, using that would most likely pose some architectural issues since constraints have the implicit contract of having two datasets when the `test` method is invoked. And in this case, we'd be receiving just one. Anyhow, that's something we can think and talk about once it's really necessary.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,47 @@
...
+
+class KolmogorovSmirnov2Sample(Constraint):
+    def __init__(
+        self, ref: DataReference, ref2: DataReference, significance_level: float = 0.05
+    ):
+        self.significance_level = significance_level
+        super().__init__(ref, ref2=ref2)
+
+    @staticmethod
+    def calculate_2sample_ks_test(data, data2):
```
ivergara, line 19:
```suggestion
    def _calculate_2sample_ks_test(data, data2) -> float:
```

You might want to try to annotate `data` and `data2`, but don't think it's critical.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        For two given lists of values calculates the Kolmogorov-Smirnov test.
+        Read more here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html
+        """
+
+        # calculate statistic
```
ivergara, line 25: I'd say that this comment is rather superfluous as it is now. Maybe you could add on why taking this approach and some of its potential failings. And even going further, mentioning at least the first level improvement of letting the engine do the CDF.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def retrieve(
+        self, engine: sa.engine.Engine, ref: DataReference
+    ) -> Tuple[Any, OptionalSelections]:
+        return db_access.get_column(engine, ref)
+
+    def compare(
+        self, value_factual: Any, value_target: Any
+    ) -> Tuple[bool, Optional[str]]:
+        # factual and target values are the corresponding columns from our data source
```
ivergara, line 38: Again, rather superfluous as comment. Seems more to clarify yourself than to help a reader of the future.

src/datajudge/db_access.py
```diff
@@ -632,43 +632,63 @@ def get_row_count(engine, ref, row_limit: int = None):
     return result, [selection]
 
 
-def _column_aggregate(engine, ref, column_operator):
+def _column(
```
ivergara, line 635: I'd maybe suggest renaming this to `get_column` and get rid of the currently existing `get_column`. It's providing an indirection that's not adding anything since the callable has a default value.

src/datajudge/db_access.py
```diff
@@ -632,43 +632,63 @@ def get_row_count(engine, ref, row_limit: int = None):
...
+    engine: sa.engine.Engine,
+    ref: DataReference,
+    aggregate_operator: Callable | None = None,
```
ivergara, line 638:
```suggestion
    *, aggregate_operator: Callable | None = None,
```
Maybe it'd be good to make this argument keyword-only, which doesn't involve other changes since it's being called always explicitly in all the convenience `get_xx` functions.

src/datajudge/requirements.py
```diff
@@ -1251,3 +1252,24 @@ def add_row_matching_equality_constraint(
...
+        column1: str,
+        column2: str,
+        condition1: Condition = None,
+        condition2: Condition = None,
+        significance_level: float = 0.05,
+    ):
+        """
+        Apply the so-called two-sample Kolmogorov-Smirnov test to the distributions of the two given columns.
+        The constraint is fulfilled, when the resulting p-value of the test is higher than the significance level
+        (default is 0.05 ie 5%).
```
ivergara, line 1267:
```suggestion
        (default is 0.05, i.e., 5%).
```

ivergara (commented):
Good job!

Take a look at the few extra observations.

src/datajudge/constraints/stats.py
YasinTatarQC: I have added numpy.typing.ArrayLike, though I'm not sure if that's the best approach as I haven't seen it being used in many other places. Makes sense, though, since that is exactly what the scipy method expects (when viewing the docs)

src/datajudge/db_access.py
YasinTatarQC: I think at this point you may be right. In the future, we might need to add back this indirection if we have more complicated data access patterns, but for now, this additional step is unnecessary perhaps.

ivergara: Sure, once we have the need this could be changed.

src/datajudge/constraints/stats.py
```diff
@@ -1,6 +1,7 @@
 from typing import Any, Optional, Tuple
 
 import sqlalchemy as sa
+from numpy import typing
```
ivergara, line 4: Since this is only needed for typing purposes better wrap it such that doesn't collide with environments not having `numpy`

```python
from typing import TYPE_CHECKING:
    if TYPE_CHECKING:
        from numpy import typing
```

src/datajudge/constraints/stats.py
ivergara: I agree that it's in principle a good approach. My only concern is what would happen when someone wants to execute any other constraint in `datajudge` without having `numpy` installed. Will the declaration of the types that don't exist blow up at runtime?

My guess is no, given the `if` guard I suggested for the import. But I have no 100% certainty. What I'm saying is that perhaps you can execute some constraints while removing `numpy` from the environment.

src/datajudge/constraints/stats.py
```diff
@@ -15,7 +16,9 @@ def __init__(
         super().__init__(ref, ref2=ref2)
 
     @staticmethod
-    def calculate_2sample_ks_test(data, data2):
+    def calculate_2sample_ks_test(
+        data: typing.ArrayLike, data2: typing.ArrayLike
```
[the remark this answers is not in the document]
YasinTatarQC: Following the guideline of being as generic as possible for inputs, I'd use `Collection` here. A quick look at the scipy source confirms that this should be fine.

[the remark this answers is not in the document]
ivergara: I hoped so too, `scipy` is very lenient on what it accepts as inputs.

src/datajudge/constraints/stats.py
YasinTatarQC: I agree with you on all points but I don't think it's worth the effort. We are perfectly fine using Python's `Iterable` or `Collection` type hints.

src/datajudge/constraints/stats.py
```diff
@@ -0,0 +1,57 @@
...
+        super().__init__(ref, ref2=ref2)
+
+    @staticmethod
+    def calculate_2sample_ks_test(data: Collection, data2: Collection) -> float:
+        """
+        For two given lists of values calculates the Kolmogorov-Smirnov test.
+        Read more here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html
+        """
+        try:
+            from scipy.stats import ks_2samp
```
kklein, line 24: I don't perfectly understand why we need this even though `scipy` is listed as dependency in `pyproject.toml`. However, if you feel confident about this, please go ahead.

tests/integration/test_integration.py
```diff
@@ -1755,3 +1755,55 @@ def test_groupby_aggregation_within_with_failures(
...
+    """
+    Test kolmogorov smirnov test for table and square of table -> significance level should be less than default 0.05
+    """
+    (operation, col_1, col_2, min_p_value) = data
+    req = requirements.BetweenRequirement.from_tables(*int_table1, *int_square_table)
+    req.add_ks_2sample_constraint(
+        column1=col_1, column2=col_2, significance_level=min_p_value
+    )
+
+    # test should fail, i.e. be below the default significance level
```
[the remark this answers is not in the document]
kklein: I think this is outdated.

kklein (approved):
Great work - thanks a lot! I'm sure this will be a very valuable addition to Datajudge.

Before merging this, it would be cool to add a changelog entry akin to the following

```
1.1.0 - 2022-xx-xx
-------------------

**New feature:**

- Implemented a :class:`~datajudge.BetweenRequirement` constraint running a Kolmogorov Smirnov test.

```

src/datajudge/constraints/stats.py
ivergara: You're right @kklein , however, the reason why `scipy` is in the `pyproject.toml` file is due to read the docs. Don't ask me why it was failing, but after adding `scipy` here it worked again.

If that weren't the case, this piece of code would be useful to manage the message the user encounters when trying to use this constraint while not having `scipy` installed.

ivergara:
> Great work - thanks a lot! I'm sure this will be a very valuable addition to Datajudge.
> 
> Before merging this, it would be cool to add a changelog entry akin to the following
> 
> ```
> 1.1.0 - 2022-xx-xx
> -------------------
> 
> **New feature:**
> 
> - Implemented a :class:`~datajudge.BetweenRequirement` constraint running a Kolmogorov Smirnov test.
> ```

You can be a bit more verbose than what Kevin suggested.

Also modify the version number in [here](https://github.com/Quantco/datajudge/blob/6be218f1ffec9218ed79567a3e0196bda93fc76c/src/datajudge/__init__.py#L22).

Once this is merged, I'll push the new version in PyPI, and given the addition of `scipy` I might have to go and change the recipe in conda-forge.

ivergara (approved):

CHANGELOG.rst
```diff
@@ -7,6 +7,17 @@
...
 =========
 
+1.1.0 - 2022.06.01
+------------------
+
+**New feature: Statistical Tests**
+
+- Implemented a :class:`~datajudge.BetweenRequirement` constraint running a Kolmogorov Smirnov test.
+
+**New method: get_column**
```
ivergara, line 17: I'd omit this one. It's an internal detail of the library and not part of what a user of this library needs in order to make data tests.

CHANGELOG.rst
```diff
@@ -7,6 +7,17 @@
 Changelog
 =========
 
+1.1.0 - 2022.06.01
+------------------
+
+**New feature: Statistical Tests**
+
+- Implemented a :class:`~datajudge.BetweenRequirement` constraint running a Kolmogorov Smirnov test.
```
ivergara, line 15:
```suggestion
- Implemented a new constraint :class:`~datajudge...` for :class:`~datajudge.BetweenRequirement` that performs a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) between two data sources.
```

You'd still need to fix the class reference for the constraint.

ivergara (changes requested):

CHANGELOG.rst
YasinTatarQC: Oh okay, I thought the changelog is also directed towards possible developers wanting to extend the library -- either officially or with their own constraints.

YasinTatarQC: I thought the titles `**title**` are splitting sections and that made it clear. I'm new to `.rst` files, is there any convention we're following or I can lookup?
At first glance, I found [this](https://www.ros.org/reps/rep-0132.html), which is, of course, from a different domain but was nevertheless referenced in a few places.

YasinTatarQC: Regarding conversation resolving: Sorry! I was cleaning up because I was losing track of some of the tasks. Makes sense of course; basic PR etiquette.

ivergara: No convention. And indeed you were having sections, but I meant not to just say "new features" or "new method" but to split into a user relevant section and a developer relevant section.

ivergara (approved):

YasinTatarQC:
Thank you both for the great feedback, learned a lot.

... see you on the next PR. 🎉

PR #27, opened 2022-06-14
Fast Data Retrieval for Analytics Databases

YYYasin19:
Hey there, 

after introducing a statistical test with #23, I now want to integrate a way to load a large batch of data (`> 100,000,000`) within a reasonable amount of time and memory use.

For this, snowflake's sqlalchemy connector already provides a convenient method that works surprisingly well. You can read more about it [here](https://docs.snowflake.com/en/user-guide/python-connector-api.html#fetch_pandas_all).

Since I dislike or am unsure about some changes here, this will be a draft first. Concretely speaking, the following things are unclear now:

- [x] Is there a way to _not_ use pandas as an intermediate representation? (Additional dependency, data needs to be transformed, etc.)
- [x] Can we only do this for snowflake, or do we have similar methods for other database systems as well?

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -12,3 +12,14 @@ or from a conda environment
 ::
 
     conda install datajudge -c conda-forge
+
+
+
+Snowflake
+^^^^
+
+If your backend is ``snowflake`` and you are querying large datasets,
+you can additionally install ``pandas`` to make use of very fast query loading
+(up to 50x speedup for large datasets).
+    Note: The ``pandas`` requirement is a bug in the snowflake-python-connector
+    and will not be needed in the future.
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import functools
-import importlib
 import json
 import operator
 from abc import ABC, abstractmethod
@@ -13,8 +12,12 @@ import sqlalchemy as sa
 from sqlalchemy.sql.expression import FromClause
 
 try:
-    snowflake_available = importlib.import_module("snowflake") is not None
-    pandas_available = importlib.import_module("pandas") is not None
+    import snowflake
+
+    snowflake_available = True
+    import pandas
+
+    pandas_available = True
 except ModuleNotFoundError as err:  # ex.: 'snowflake' not found
     snowflake_available = "snowflake" not in str(err)
     pandas_available = "pandas" not in str(err)
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -654,7 +654,7 @@ def get_column(
         if is_snowflake(engine):
             snowflake_cursor: SnowflakeCursor = engine.connect().connection.cursor()
 
-            # note: this step requires pandas to be installed
+            # note: in addition to pyarrow, this currently requires pandas as well
             pa_table = snowflake_cursor.execute(str(selection)).fetch_arrow_all()
             if pa_table:  # snowflake connector returns NoneType when the table is empty
                 result = pa_table.column(0).to_numpy()
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -11,7 +11,8 @@ from typing import Callable, Sequence, final, overload
 
 import sqlalchemy as sa
 from sqlalchemy.sql.expression import FromClause
-from utils import check_module_installed
+
+from .utils import check_module_installed
 
 snowflake_available = check_module_installed("snowflake")
 pandas_available = check_module_installed("pandas")
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -16,6 +16,9 @@ try:
     pandas_available = importlib.import_module("pandas") is not None
 except ModuleNotFoundError:
     pandas_available = False
+    print(
+        "For snowflake users: `pandas` is not installed, that means optimized data loading is not available."
+    )
 
 
 def is_mssql(engine: sa.engine.Engine) -> bool:
@@ -655,7 +658,7 @@ def get_column(
     if not aggregate_operator:
         selection = sa.select([column])
 
-        # snowflake-specific optimization iff pandas is installed additionally
+        # snowflake-specific optimization iff pandas is installed
         if is_snowflake(engine) and pandas_available:
             snowflake_cursor = engine.connect().connection.cursor()
 
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -9,7 +9,6 @@ from dataclasses import dataclass
 from typing import Callable, Sequence, final, overload
 
 import sqlalchemy as sa
-from snowflake.connector.cursor import SnowflakeCursor
 from sqlalchemy.sql.expression import FromClause
 
 
@@ -652,7 +651,7 @@ def get_column(
 
         # snowflake-specific optimization
         if is_snowflake(engine):
-            snowflake_cursor: SnowflakeCursor = engine.connect().connection.cursor()
+            snowflake_cursor = engine.connect().connection.cursor()
 
             # note: in addition to pyarrow, this currently requires pandas as well
             pa_table = snowflake_cursor.execute(str(selection)).fetch_arrow_all()
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import functools
 import json
 import operator
+import warnings
 from abc import ABC, abstractmethod
 from collections import Counter
 from dataclasses import dataclass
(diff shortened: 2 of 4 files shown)


YYYasin19:
> My guess is that we should revolve around arrow tables and use the following [method](https://docs.snowflake.com/en/user-guide/python-connector-api.html#fetch_arrow_all) instead. [This](https://arrow.apache.org/docs/python/numpy.html) could also help.

Yes, that could also make sense, I thought it wouldn't really make a difference since we'd need a new requirement for both of them and, in my experience, most projects already have `pandas` installed for some reason. 
Or is Arrow already used somewhere within the project?

ivergara:
> > My guess is that we should revolve around arrow tables and use the following [method](https://docs.snowflake.com/en/user-guide/python-connector-api.html#fetch_arrow_all) instead. [This](https://arrow.apache.org/docs/python/numpy.html) could also help.
> 
> Yes, that could also make sense, I thought it wouldn't really make a difference since we'd need a new requirement for both of them and, in my experience, most projects already have `pandas` installed for some reason. Or is Arrow already used somewhere within the project?

pyarrow is already installed by `snowflake-connector`

https://github.com/snowflakedb/snowflake-connector-python/blob/2ccd3c811958170854ff4cac159e7c125bf8b4f3/pyproject.toml#L2-L10

In a summary, arrow intends to be the common in-memory representation of tables/dataframes enabling interoperability between different analytics applications. It has plenty of other advantages, that the `snowflake-connector is using for example to improve data transfer.

And I'm more comfortable going that route than straight out requiring pandas. pandas, for what you want to do, is only an intermediary... and a rather bad one at that in my opinion.

YYYasin19:
> pyarrow is already installed by `snowflake-connector`

Oh, didn't catch that! Then it makes perfect sense to just build on arrow. All of our snowflake users won't have to have anything additionally installed then.

src/datajudge/db_access.py
```diff
@@ -9,6 +9,7 @@
 from typing import Callable, Sequence, final, overload
 
 import sqlalchemy as sa
+from snowflake.connector.cursor import SnowflakeCursor
```
ivergara, line 12: Guard this import
```python
try:
    from...
except ImportError:
    print("a useful message")
```

src/datajudge/db_access.py
```diff
@@ -650,14 +651,12 @@ def get_column(
         selection = sa.select([column])
 
         if is_snowflake(engine):  # check if we have a snowflake cursor
-            snowflake_cursor = engine.connect().connection.cursor()
+            snowflake_cursor: SnowflakeCursor = engine.connect().connection.cursor()
 
             # note: this step requires pandas to be installed
```
ivergara, line 657: This comment is outdated I believe.

YYYasin19:
Unfortunately, the snowflake connector is causing a segmentation fault when pandas is _not_ installed.
The segmentation fault occurs in the `snowflake-python-connector` code, precisely [when yielding from the iterator within the ThreadPoolExecutor](https://github.com/snowflakedb/snowflake-connector-python/blob/2ccd3c811958170854ff4cac159e7c125bf8b4f3/src/snowflake/connector/result_set.py#L68).

I'll investigate this further.

ivergara:
> Unfortunately, the snowflake connector is causing a segmentation fault when pandas is _not_ installed. The segmentation fault occurs in the `snowflake-python-connector` code, precisely [when yielding from the iterator within the ThreadPoolExecutor](https://github.com/snowflakedb/snowflake-connector-python/blob/2ccd3c811958170854ff4cac159e7c125bf8b4f3/src/snowflake/connector/result_set.py#L68).

Not much info on segmentation faults in `snowflake-python-connector`, [this](https://github.com/snowflakedb/snowflake-connector-python/issues/1100) might be the closest I found, which it's "solved" by installing `pandas`

> I'll investigate this further.

👍

src/datajudge/db_access.py
jonashaag: Btw nowadays we have `ModuleNotFoundError`

src/datajudge/db_access.py
ivergara: yes

YYYasin19: Is it, though? Because currently, it requires `pyarrow` _and_ `pandas` because we get the SegFault otherwise. I'll change it to reflect that, though.

src/datajudge/db_access.py
YYYasin19: I removed the import statement since it only is a type hint for the user

src/datajudge/db_access.py
ivergara: I wrote it before I was aware of the segfault without pandas.

src/datajudge/db_access.py
ivergara: Then you should do
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ....
```

src/datajudge/db_access.py
```diff
@@ -648,7 +649,19 @@ def get_column(
...
         selection = sa.select([column])
-        result = engine.connect().execute(selection).scalars().all()
+
+        if is_snowflake(engine):  # check if we have a snowflake cursor
+            snowflake_cursor: SnowflakeCursor = engine.connect().connection.cursor()
+
+            # note: this step requires pandas to be installed
+            pa_table = snowflake_cursor.execute(str(selection)).fetch_arrow_all()
+            if pa_table:  # snowflake connector returns NoneType when the table is empty
+                result = pa_table.column(0).to_numpy()
```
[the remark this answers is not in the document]
ivergara: > Just accessing the first column looks unsafe, but should ideally be always correct, since our `selection` only selects a single column.

That's fine.

src/datajudge/db_access.py
YYYasin19: I don't think I really get what you mean.

I initially added the import to have a 
```py
cursor: SnowflakeCursor = engine.connect().connection.cursor()
```
line, where the IDE then _knows_ that cursor has additional methods provided by snowflake.

Since this was not needed anymore (not important enough) I removed it again, so it had no checking functionality. So it should be fine to just remove it.

YYYasin19:
> If this experiment gets merged, we have to add a note in the documentation concerning Snowflake use the need of manually installing `pandas`.

I think you are right and this is the best way currently. 
I have added a warning message for snowflake users, though, that when pandas is not available, the user is aware that optimizations won't be run.
The only issue is that I don't know how to show the message to snowflake-users only.

YYYasin19:
Also, the issue is now open [here](https://github.com/snowflakedb/snowflake-connector-python/issues/1169) in the `snowflake-python-connector`.

src/datajudge/db_access.py
```diff
@@ -648,7 +657,20 @@ def get_column(
 
     if not aggregate_operator:
         selection = sa.select([column])
-        result = engine.connect().execute(selection).scalars().all()
+
+        # snowflake-specific optimization iff pandas is installed
+        if is_snowflake(engine) and pandas_available:
```
ivergara, line 662:
```suggestion
        if is_snowflake(engine):
            if not pandas_available:
                print message and exit the snowflake code returning to sqlalchemy path.
```

Concerning the message, this is an option. You delay using the flag you created until here.

YYYasin19: I wanted to try that one as well, unfortunately, now the error message would be printed for each and every query.

ivergara: Fair enough, then you can try to import snowflake, if it succeeds you can do the pandas check and inform in case it's not present.

jonashaag:
IMO I think Snowflake-only is fine

ivergara:
> The essential question at this point is: Do we implement fast(er) loading for `mssql` and `postgres` as well or are we satisfied with just `snowflake` for now; maybe until someone opens an issue for another backend?

Just Snowflake is fine, since the work is done already. For other backends, we see in the future.

src/datajudge/db_access.py
YYYasin19: @ivergara are you satisfied with this solution?

docs/source/installation.rst
```diff
@@ -12,3 +12,14 @@ or from a conda environment
...
+
+
+Snowflake
+^^^^
+
+If your backend is ``snowflake`` and you are querying large datasets,
+you can additionally install ``pandas`` to make use of very fast query loading
+(up to 50x speedup for large datasets).
+    Note: The ``pandas`` requirement is a bug in the snowflake-python-connector
+    and will not be needed in the future.
```
ivergara, line 25:
```suggestion
    and, hopefully, it will not be needed in the future.
```

src/datajudge/utils.py
```diff
@@ -0,0 +1,8 @@
+def check_module_installed(module_name: str) -> bool:
+    import importlib
+
+    try:
+        mod = importlib.import_module(module_name)
+        return mod is not None
```
ivergara, line 6: Why not just return `True`? Or can it happen that the module is not found and still not trigger the exception?

src/datajudge/db_access.py
ivergara: Sure.

ivergara (commented):
Looks good to me. What's your opinion @kklein?

src/datajudge/utils.py
YYYasin19: Nope, just checked, it always throws an error :)

ivergara:
Since the main motivation to go for this solution has been solved differently (see #28), there is no pressing need of using native Snowflake communication. Thus, this PR is currently out of scope for the project and will be closed.

Thank you for your effort @YYYasin19. Maybe in the future, the needs change and we'll find ourselves needing to do something that cannot be pushed to the engine and this exploration might come in handy.

PR #28, opened 2022-06-24
Implement Kolmogorov Smirnov Test in SQL-only

YYYasin19:
Hey!

in this PR I'd like to refactor the work of #23 and now implement the Kolmogorov-Smirnov test in SQL-only.
This has the benefit that for large datasets, we won't have to load in all the values to pass them to scipy.
Additionally, this would remove scipy as a dependency again! 🥳 

The core work of this PR consists of creating an SQL query that is able to calculate the [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test#Two-sample_Kolmogorov.E2.80.93Smirnov_test). 

The idea behind this implementation as well as some theoretical backgrounds will be explained in the comments below. 

The following issues will definitely need to be addressed:

- [x] Create more tests, especially with very large datasets
- [x] Refactor tests as p_values are not always available now
- [x] ~Remove scipy dependency~ needs to be done on conda-feedstock
- [x] Test if the query is really database-agnostic, i.e., follows some SQL standard 
- [x] ~Transform query to `sqlalchemy`'s query API~ #29 issue opened
- [x] Validate p-value approximation

### DIFF
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -64,6 +64,19 @@ class KolmogorovSmirnov2Sample(Constraint):
             (n_samples + m_samples) / (n_samples * m_samples)
         )
 
+    @staticmethod
+    def calculate_statistic(engine, table1, table2) -> Any:
+
+        # retrieve test statistic d, as well as sample sizes m and n
+        d_statistic, m, n = db_access.get_ks_2sample(
+            engine, table1=table1, table2=table2
+        )
+
+        # calculate approximate p-value
+        p_value = KolmogorovSmirnov2Sample.approximate_p_value(d_statistic, m, n)
+
+        return d_statistic, p_value, n, m
+
     def test(self, engine: sa.engine.Engine) -> TestResult:
 
         # get query selections and column names for target columns
@@ -72,16 +85,14 @@ class KolmogorovSmirnov2Sample(Constraint):
         selection2 = str(self.ref2.data_source.get_clause(engine))
         column2 = self.ref2.get_column(engine)
 
-        # retrieve test statistic d, as well as sample sizes m and n
-        d_statistic, m, n = db_access.get_ks_2sample(
-            engine, table1=(selection1, column1), table2=(selection2, column2)
+        d_statistic, p_value, n_samples, m_samples = self.calculate_statistic(
+            engine, (selection1, column1), (selection2, column2)
         )
 
-        # calculate approximate p-value
-        p_value = self.approximate_p_value(d_statistic, m, n)
-
         # calculate test acceptance
-        result = self.check_acceptance(d_statistic, n, m, self.significance_level)
+        result = self.check_acceptance(
+            d_statistic, n_samples, m_samples, self.significance_level
+        )
 
         assertion_text = (
             f"Null hypothesis (H0) for the 2-sample Kolmogorov-Smirnov test was rejected, i.e., "
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,6 +1,6 @@
 import math
 import warnings
-from typing import Optional
+from typing import Optional, Union
 
 import sqlalchemy as sa
 from sqlalchemy.sql import Selectable
@@ -69,8 +69,8 @@ class KolmogorovSmirnov2Sample(Constraint):
     @staticmethod
     def calculate_statistic(
         engine,
-        table1_def: tuple[Selectable | str, str],
-        table2_def: tuple[Selectable | str, str],
+        table1_def: tuple[Union[Selectable, str], str],
+        table2_def: tuple[Union[Selectable, str], str],
     ) -> tuple[float, Optional[float], int, int]:
 
         # retrieve test statistic d, as well as sample sizes m and n
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,10 +1,12 @@
-from typing import Any, Collection, Optional, Tuple
+import math
+import warnings
+from typing import Any, Tuple
 
 import sqlalchemy as sa
 
 from .. import db_access
 from ..db_access import DataReference
-from .base import Constraint, OptionalSelections
+from .base import Constraint, OptionalSelections, TestResult
 
 
 class KolmogorovSmirnov2Sample(Constraint):
@@ -14,44 +16,66 @@ class KolmogorovSmirnov2Sample(Constraint):
         self.significance_level = significance_level
         super().__init__(ref, ref2=ref2)
 
-    @staticmethod
-    def calculate_2sample_ks_test(data: Collection, data2: Collection) -> float:
-        """
-        For two given lists of values calculates the Kolmogorov-Smirnov test.
-        Read more here: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html
-        """
-        try:
-            from scipy.stats import ks_2samp
-        except ModuleNotFoundError:
-            raise ModuleNotFoundError(
-                "Calculating the Kolmogorov-Smirnov test relies on scipy."
-                "Therefore, please install scipy before using this test."
-            )
-
-        # Currently, the calculation will be performed locally through scipy
-        # In future versions, an implementation where either the database engine
-        # (1) calculates the CDF
-        # or even (2) calculates the KS test
-        # can be expected
-        statistic, p_value = ks_2samp(data, data2)
-
-        return p_value
-
     def retrieve(
         self, engine: sa.engine.Engine, ref: DataReference
     ) -> Tuple[Any, OptionalSelections]:
-        return db_access.get_column(engine, ref)
+        sel = ref.get_selection(engine)  # table selection incl. WHERE condition
+        col = ref.get_column(engine)  # column name
+        return sel, col
 
-    def compare(
-        self, value_factual: Any, value_target: Any
-    ) -> Tuple[bool, Optional[str]]:
+    @staticmethod
+    def approximate_p_value(d: float, m: int, n: int) -> float | None:
+        """
+        Calculates the approximate p-value according to
+        'A procedure to find exact critical values of Kolmogorov-Smirnov Test', Silvia Fachinetti, 2009
+        """
 
-        p_value = self.calculate_2sample_ks_test(value_factual, value_target)
-        result = p_value >= self.significance_level
-        assertion_text = (
-            f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
-            f"has p-value {p_value}  < {self.significance_level}"
-            f"{self.condition_string}"
+        n = max(m, n) // min(m, n)
+        if n < 35:
+            warnings.warn(
+                "Approximating the p-value is not accurate enough for sample size < 35"
+            )
+            return None
+
+        d_alpha = d * math.sqrt(n)
+        return -2 * math.exp(-(d_alpha**2))
+
+    @staticmethod
+    def check_acceptance(d: float, n: int, m: int, accepted_level):
+        def c(alpha: float):
+            return math.sqrt(-math.log(alpha / 2.0) * 0.5)
+
+        # source: https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test
+        return d <= c(accepted_level) * math.sqrt((n + m) / (n * m))
+
+    def test(self, engine: sa.engine.Engine) -> TestResult:
+
+        # get query selections and column names for target columns
+        selection1 = self.ref.get_selection(engine)
+        column1 = self.ref.get_column(engine)
+        selection2 = self.ref2.get_selection(engine)
+        column2 = self.ref2.get_column(engine)
+
+        # retrieve test statistic d, as well as sample sizes m and n
+        d_statistic, m, n = db_access.get_ks_2sample(
+            engine, table1=(selection1, column1), table2=(selection2, column2)
         )
 
-        return result, assertion_text
+        # calculate approximate p-value
+        p_value = self.approximate_p_value(d_statistic, m, n)
+
+        # calculate test acceptance
+        result = self.check_acceptance(d_statistic, n, m, self.significance_level)
+
+        assertion_text = (
+            f"Null hypothesis (H0) for the 2-sample Kolmogorov-Smirnov test was rejected, i.e., "
+            f"the two samples ({self.ref.get_string()} and {self.target_prefix})"
+            f" do not originate from the same distribution."
+        )
+        if p_value:
+            assertion_text += f"\n p-value: {p_value}"
+
+        if not result:
+            return TestResult.failure(assertion_text)
+
+        return TestResult.success()
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,12 +1,12 @@
 import math
 import warnings
-from typing import Any, Optional, Tuple
+from typing import Any, Optional
 
 import sqlalchemy as sa
 
 from .. import db_access
 from ..db_access import DataReference
-from .base import Constraint, OptionalSelections, TestResult
+from .base import Constraint, TestResult
 
 
 class KolmogorovSmirnov2Sample(Constraint):
@@ -15,13 +15,6 @@ class KolmogorovSmirnov2Sample(Constraint):
     ):
         self.significance_level = significance_level
         super().__init__(ref, ref2=ref2)
-
-    def retrieve(
-        self, engine: sa.engine.Engine, ref: DataReference
-    ) -> Tuple[Any, OptionalSelections]:
-        sel = ref.get_selection(engine)  # table selection incl. WHERE condition
-        col = ref.get_column(engine)  # column name
-        return sel, col
 
     @staticmethod
     def approximate_p_value(
@@ -56,12 +49,12 @@ class KolmogorovSmirnov2Sample(Constraint):
     @staticmethod
     def check_acceptance(
         d_statistic: float, n_samples: int, m_samples: int, accepted_level: float
-    ):
+    ) -> bool:
         """
         For a given test statistic, d, and the respective sample sizes `n` and `m`, this function
         checks whether the null hypothesis can be rejected for an accepted significance level.
 
-        For more information, check out the `Wikipedia entry <https://w.wiki/5May>`.
+        For more information, check out the `Wikipedia entry <https://w.wiki/5May>`_.
         """
 
         def c(alpha: float):
@@ -72,24 +65,26 @@ class KolmogorovSmirnov2Sample(Constraint):
         )
 
     @staticmethod
-    def calculate_statistic(engine, table1, table2) -> Any:
+    def calculate_statistic(engine, table1_def: tuple, table2_def: tuple) -> Any:
 
         # retrieve test statistic d, as well as sample sizes m and n
-        d_statistic, m, n = db_access.get_ks_2sample(
-            engine, table1=table1, table2=table2
+        d_statistic, n_samples, m_samples = db_access.get_ks_2sample(
+            engine, table1=table1_def, table2=table2_def
         )
 
         # calculate approximate p-value
-        p_value = KolmogorovSmirnov2Sample.approximate_p_value(d_statistic, m, n)
+        p_value = KolmogorovSmirnov2Sample.approximate_p_value(
+            d_statistic, n_samples, m_samples
+        )
 
-        return d_statistic, p_value, n, m
+        return d_statistic, p_value, n_samples, m_samples
 
     def test(self, engine: sa.engine.Engine) -> TestResult:
 
         # get query selections and column names for target columns
-        selection1 = str(self.ref.data_source.get_clause(engine))
+        selection1 = self.ref.data_source.get_clause(engine)
         column1 = self.ref.get_column(engine)
-        selection2 = str(self.ref2.data_source.get_clause(engine))
+        selection2 = self.ref2.data_source.get_clause(engine)
         column2 = self.ref2.get_column(engine)
 
         d_statistic, p_value, n_samples, m_samples = self.calculate_statistic(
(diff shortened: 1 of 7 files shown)


YYYasin19:
## Kolmogorov Smirnov in SQL

Our sample table `table1` has the following CDF
 VAL | CDF          
-----|--------------
 1   | 0.2352941176 
 2   | 0.3529411765 
 3   | 0.4705882353 
 4   | 0.5882352941 
 5   | 0.7058823529 
 6   | 0.8235294118 
 7   | 0.9411764706 
 8   | 1            

which is calculated by the first query and stored in `tab1`. The same applies to `tab2`.
Since the range of the values in `table1` might differ from the range of values in `table2`, we perform a `FULL OUTER JOIN` between the CDFs based on their index (here: `val`, i.e., the x-axis).


```sql
SELECT coalesce(tab1.val, tab2.val) as v, tab1.cdf as cdf1, tab2.cdf as cdf2
FROM tab1 FULL OUTER JOIN tab2 ON tab1.val = tab2.val
ORDER BY v
```
Using `coalesce` we combine the indices, which results in the following table:
| V  | CDF1         | CDF2         |
|----|--------------|--------------|
| 1  | 0.1666666667 | 0.2352941176 |
| 2  | 0.2222222222 | 0.3529411765 |
| 3  | 0.2777777778 | 0.4705882353 |
| 4  | 0.3333333333 | 0.5882352941 |
| 5  | 0.3888888889 | 0.7058823529 |
| 6  | 0.4444444444 | 0.8235294118 |
| 7  | 0.5          | 0.9411764706 |
| 8  |              | 1            |
| 16 | 0.5555555556 |              |
| 17 | 0.6111111111 |              |
| 18 | 0.6666666667 |              |
| 19 | 0.7222222222 |              |
| 20 | 0.7777777778 |              |
| 21 | 0.8333333333 |              |
| 22 | 0.8888888889 |              |
| 23 | 0.9444444444 |              |
| 24 | 1            |              |


### Filling missing values
Since our discrete values result in a step-function like CDF, we want to forward-fill missing values in the table.
This is done by first assigning a grouper_id to each (index, cdf_value) combination with
```sql
grouped_table AS ( -- Step 2: Create a grouper attribute to fill values forward
    SELECT v,
        COUNT(cdf1) over (order by v) as _grp1, cdf1,
        COUNT(cdf2) over (order by v) as _grp2, cdf2
    FROM cdf_unfilled
```
and then retrieving the first value per group here
```sql
filled_cdf AS ( -- Step 3: Select first non-null value per group (defined by grouper)
    SELECT v, first_value(cdf1) over (partition by _grp1 order by v) as cdf1_filled,
            first_value(cdf2) over (partition by _grp2 order by v) as cdf2_filled
    FROM grouped_table
```

### Replacing NULLs
As long as the smallest value of both data samples is not the same, e.g., column1 starts with 1 and column2 starts with 5, one of the CDF tables will have NULL as first value. The following query replaces those resulting NULL values with 0, s.t. the distance can be effectively calculated.


### Final Query

The final query has, currently, the following form.
Some of the queries can still be optimized. The current structure allows us to just insert datajudge's `get_clause` into the query string to have a filtered table and use a `val` alias on the target column.

```sql
WITH tab1 AS (
  SELECT val, MAX(cdf) as cdf FROM (
      SELECT val, cume_dist() over (order by val) as cdf
      FROM (SELECT {column1} as val FROM {table1}) -- Change TABLE and COL here
      ORDER BY val
  ) GROUP BY val

), tab2 AS (
  SELECT val, MAX(cdf) as cdf FROM (
      SELECT val, cume_dist() over (order by val) as cdf
      FROM (SELECT {column2} as val FROM {table2}) -- Change TABLE and COL here
      ORDER BY val
  ) GROUP BY val

), cdf_unfilled AS (
  SELECT coalesce(tab1.val, tab2.val) as v, tab1.cdf as cdf1, tab2.cdf as cdf2
  FROM tab1 FULL OUTER JOIN tab2 ON tab1.val = tab2.val
  ORDER BY v

), grouped_table AS ( -- Step 2: Create a grouper attribute to fill values forward
    SELECT v,
        COUNT(cdf1) over (order by v) as _grp1, cdf1,
        COUNT(cdf2) over (order by v) as _grp2, cdf2
    FROM cdf_unfilled

), filled_cdf AS ( -- Step 3: Select first non-null value per group (defined by grouper)
    SELECT v, first_value(cdf1) over (partition by _grp1 order by v) as cdf1_filled,
            first_value(cdf2) over (partition by _grp2 order by v) as cdf2_filled
    FROM grouped_table

), replaced_nulls AS ( -- Step 4: Replace NULL values (often at the begin) with 0 to calculate difference
  SELECT coalesce(cdf1_filled, 0) as cdf1, coalesce(cdf2_filled, 0) as cdf2
  FROM filled_cdf
  )

SELECT MAX(ABS(cdf1-cdf2)) FROM replaced_nulls; -- Step 5: Calculate final statistic
```

YYYasin19:
## Theoretical foundation for p-value
The query above only calculates the statistic of the Kolmogorov Smirnov test, which is defined as
$$\left[ \sup_{x} || F_1(x) - F_2(x) || \right] $$
where $F_1(x)$ and $F_2(x)$ are the CDFs of the two samples $s_1$ and $s_2$ respectively.


### Determining acceptance
We can now accept or reject the null-hypothesis $H_0$ of our test based on our test statistic $d$ and our sample sizes $n$ and $m$ with the following inequality:
$$ d \lt \sqrt{-\ln{\frac{\alpha}{2} * \frac{1}{2}}} $$
In this case, $H_0$ states that the two samples stem from the same distribution.

### Approximating the p-value
Since this value is test-specific and harder to interpret across applications, it would be beneficial to have a p-value at hand!
Unfortunately, the distribution of the KS-statistic is unwieldly, making exact calculations hard. [(Facchinetti, 2007)](http://sa-ijas.stat.unipd.it/sites/sa-ijas.stat.unipd.it/files/IJAS_3-4_2009_07_Facchinetti.pdf) describes the following formula as an approximation of the p-value given $d$ and sample size $n$ for samples with $n \gt 35$:
$$ 2*\exp^{-(d*\sqrt(n))^2} $$

After experimenting with a few test datasets, both very small (~50 samples) and very large (~100,000,000 samples), I can confirm that this approximation is mostly correct for the first few decimal places. (Very unscientific, I know).

### Implementation:
For samples with $n \lt 35$ we won't provide a p-value for now. Also accepting or rejecting the test will still be based on the above inequality, the approx. p-value is solely for reporting purposes.

ivergara:
Great job!

> The following issues will definitely need to be addressed:
> 
> * [ ]  Create more tests, especially with very large datasets

I'm not sure how to deal with tests of large datasets. Maybe those will need to be executed manually on demand.

> * [ ]  Test if the query is really database-agnostic, i.e., follows some SQL standard

Well, if you write the tests and it passes on the three databases we test against, then you're done.

> * [ ]  Transform query to `sqlalchemy`'s query API

I don't think it'll be very feasible or advisable to attempt that. ORMs have some limitations and sometimes going straight to SQL is necessary and advisable. 

Anyhow, besides some minor issues that I'll point out later when this PR leaves draft mode, all looks good. Perhaps try to be more consistent in the type annotations including return values.

kklein:
> I don't think it'll be very feasible or advisable to attempt that. ORMs have some limitations and sometimes going straight to SQL is necessary and advisable.

Maybe you @YYYasin19 meant the sqlalchemy language expression api? We use that in the rest of `db_access`. We don't use the ORM api.

ivergara:
> > I don't think it'll be very feasible or advisable to attempt that. ORMs have some limitations and sometimes going straight to SQL is necessary and advisable.
> 
> Maybe @YYYasin19 meant the sqlalchemy language expression api? We use that in the rest of `db_access`. We don't use the ORM api.

Duh, sorry. Still, in this case, I'd advise against it. I'm not opposed to it, just that don't think it's worth the effort.

YYYasin19:
Is there a reason why our MSSQL code is converting the table specification in `ref.get_selection()` into a string with quotation marks, like this `"tempdb.dbo".int_table1.col_int `?
This breaks the query for now, other than that, everything seems to work 🥳

kklein:
> why our MSSQL code is converting the table specification in ref.get_selection() into a string with quotation marks, like this

I'm not perfectly sure - I would've thought that `get_selection` returns a selection rather than a string but I might be missing something here.

Independently of that, I know that we sometimes ran into trouble with mssql when it came to referring to objects. E.g.:

> Table names can contain any valid characters (for example, spaces). If table names contain any characters except letters, numbers, and underscores, the name must be delimited by enclosing it in back quotes (`).

https://docs.microsoft.com/en-us/sql/odbc/microsoft/table-name-limitations?view=sql-server-ver16

YYYasin19:
> I'm not perfectly sure - I would've thought that `get_selection` returns a selection rather than a string but I might be missing something here.

Yes, it does, I just had wrapped a `str(...)` around it 😬 .
I now have a solution that is very sub-optimal as I:
(1) replace `"` in MS-SQL table definitions manually
```py
    if is_mssql(engine):
        table1_selection = str(table1_selection).replace('"', "")
        table2_selection = str(table2_selection).replace('"', "")
```
Note: `sa.select(...)` is the culprit here and returns `"tempdb.dbo".table_name` instead of `tempdb.dbo.table_name`.

(2) rely on `data_source.table_name`, which is not available for `RawQueryDataSource`s
Also, this would not include any `where`-clauses, which makes it straight up wrong in many cases.

src/datajudge/constraints/stats.py
```diff
@@ -14,44 +16,73 @@ def __init__(
...
-        return p_value
+        # clamp value to [0, 1]
+        return 1.0 if approx_p > 1.0 else 0.0 if approx_p < 0.0 else approx_p
 
-    def retrieve(
-        self, engine: sa.engine.Engine, ref: DataReference
-    ) -> Tuple[Any, OptionalSelections]:
-        return db_access.get_column(engine, ref)
+    @staticmethod
+    def check_acceptance(d: float, n: int, m: int, accepted_level):
```
jonashaag, line 47: Unless d, n, m are well-known shortcuts in statistics I'd go with proper variable names here and have a note what's the equivalent on that Wikipedia page, eg.

```
def ... sample1_size: int, sample2_size: int ...
"""
See [wikipedia url] for reference with n = sample1_size, m = sample2_size ...
"""

Comment (jonashaag) [reply src/datajudge/constraints/stats.py:47]: Also a docstring for this function would be great

File: src/datajudge/constraints/stats.py [hunk 2]
```diff
@@ -14,44 +16,73 @@ def __init__(
...
+        def c(alpha: float):
+            if alpha == 0.0:
```
jonashaag, line 49: `==` comparisons with floating point numbers are dangerous, generally speaking.

If you want to avoid log(0) I think it would be better to just do `log(alpha / 2.0 + 1e-10)`. This also covers the problem of very small `alpha`. Could add an `assert alpha > 0`.

src/datajudge/db_access.py
```diff
@@ -902,3 +902,60 @@ def get_column_array_agg(
...
+    Runs the query for the two-sample Kolmogorov-Smirnov test and returns the test statistic d.
+    """
+    table1_selection, col1 = table1
+    table2_selection, col2 = table2
+
+    if is_mssql(engine):
+        table1_selection = str(table1_selection).replace('"', "")
+        table2_selection = str(table2_selection).replace('"', "")
+
+    ks_query_string = f"""
```
jonashaag, line 918: The SQL here is fairly complicated so it would be great to have some explanation. I think you can just copy and paste your GH PR text.

src/datajudge/constraints/stats.py
ivergara: I do agree with both comments, however, since `check_acceptance` is an internal function I'm not too fuzzed about the variable naming IF you improve the docstring. However, improving the argument names would be a great benefit.. we're not in the 70's anymore, where economy of names was a concern.

YYYasin19: Haha I get your points, just didn't think about it during the initial setup since all variables in the formulas are called `n` and `m` as well. `n` is very common in statistics for the sample size.

src/datajudge/db_access.py
YYYasin19: Is it okay for you if I just link to the full explanation here and give very short in-line comments in the code?

YYYasin19:
How do I find out if my approximation of the p-value is good enough?
Left is the approximation, right is the "correct" one from scipy
0.00035221594346540835 vs 0.00035250250849476917

Is it feasible to check whether the user has scipy installed and if so, just use scipy?

tests/integration/test_integration.py
```diff
@@ -1835,43 +1842,42 @@ def test_ks_2sample_random(engine, random_normal_table, configuration):
 @pytest.mark.parametrize(
     "configuration",
     [
-        ("value_0_1", "value_0_1"),
-        ("value_0_1", "value_02_1"),
-        ("value_0_1", "value_1_1"),
+        ("value_0_1", "value_0_1", 0.0, 1.0),
```
ivergara, line 1845: Add a comment mentioning how you got the values you have in these parametrizations. Namely numpy/scipy version, compiler, platform/architecture/ Anything that might be helpful to track down possible issues in the future.

tests/integration/test_integration.py
```diff
@@ -1835,43 +1842,42 @@ def test_ks_2sample_random(engine, random_normal_table, configuration):
...
-    column1 = constraint.ref.get_column(engine)
-    selection2 = str(constraint.ref2.data_source.get_clause(engine))
-    column2 = constraint.ref2.get_column(engine)
-
-    # retrieve test statistic d, as well as sample sizes m and n
-    d_statistic, m, n = db_access.get_ks_2sample(
-        engine, table1=(selection1, column1), table2=(selection2, column2)
-    )
+    col_1, col_2, expected_d, expected_p = configuration
+    # different order because schema is optional
```
ivergara, line 1853:
```suggestion
    # different order between values from fixture and `TableDataSource` arguments
```

I'm not sure it's needed to comment on this, but it's not wrong doing it.

ivergara (commented):
Once again, good job!

ivergara:
> How do I find out if my approximation of the p-value is good enough? Left is the approximation, right is the "correct" one from scipy **0.000352**21594346540835 vs **0.000352**50250849476917 The difference is 2.8656502936081404e-07

Usually, the level of "precision" shown by calculators and computers is somewhat artificial. I'm not a statistician, but I'd say that anything any difference beyond e-06, in this case, is irrelevant. Usually one uses the p value of 0.05 or if you want to be very picky 0.01. A difference in the calculated value via SQL and scipy like you find will not make anyone accept or reject the null hypothesis.

> Is it feasible to check whether the user has scipy installed and if so, just use scipy?

Well, no... this is only executed in test. Users would never use this, so it's irrelevant if users have `scipy` installed or not.

tests/integration/test_integration.py
YYYasin19: Good idea 👍

tests/integration/test_integration.py
YYYasin19: Yeah you're right, it's not that essential. Will be removed 👍

YYYasin19:
> Usually one uses the p value of 0.05 or if you want to be very picky 0.01. A difference in the calculated value via SQL and scipy like you find will not make anyone accept or reject the null hypothesis.

The problem is that in practice for one of our tests the reported p-value in the magnitude of `1e-53` because the table has 100+ million rows. Therefore, my idea to use scipy in very scientific environments where accuracy is required.

> Well, no... this is only executed in test.
No, I mean in production. 

In the end, it would just be an extra statement within the approximation method
```py
try:
   import scipy
except ModuleNotFoundError:
   return # old method of approximation

# accurate method
return scipy.stats.distribution.kstwo.sf(d, n_samples)
```

I think that's feasible 👍

src/datajudge/constraints/stats.py
```diff
@@ -14,44 +16,94 @@ def __init__(
...
-        result = p_value >= self.significance_level
         assertion_text = (
-            f"2-Sample Kolmogorov-Smirnov between {self.ref.get_string()} and {self.target_prefix}"
-            f"has p-value {p_value}  < {self.significance_level}"
-            f"{self.condition_string}"
+            f"Null hypothesis (H0) for the 2-sample Kolmogorov-Smirnov test was rejected, i.e., "
+            f"the two samples ({self.ref.get_string()} and {self.target_prefix})"
+            f" do not originate from the same distribution."
         )
+        if p_value:
```
jonashaag, line 102:
```suggestion
        if p_value is not None:
```
?

src/datajudge/constraints/stats.py
```diff
@@ -14,44 +16,94 @@ def __init__(
...
-        return db_access.get_column(engine, ref)
+    @staticmethod
+    def check_acceptance(
+        d_statistic: float, n_samples: int, m_samples: int, accepted_level: float
+    ):
+        """
+        For a given test statistic, d, and the respective sample sizes `n` and `m`, this function
+        checks whether the null hypothesis can be rejected for an accepted significance level.
+
+        For more information, check out the `Wikipedia entry <https://w.wiki/5May>`.
```
jonashaag, line 57:
```suggestion
        For more information, check out the `Wikipedia entry <https://w.wiki/5May>_`.
```

If this is RST syntax I think you need this

src/datajudge/constraints/stats.py
```diff
@@ -14,44 +16,94 @@ def __init__(
...
-    ) -> Tuple[bool, Optional[str]]:
+        return d_statistic <= c(accepted_level) * math.sqrt(
+            (n_samples + m_samples) / (n_samples * m_samples)
+        )
+
+    @staticmethod
+    def calculate_statistic(engine, table1, table2) -> Any:
+
+        # retrieve test statistic d, as well as sample sizes m and n
+        d_statistic, m, n = db_access.get_ks_2sample(
```
jonashaag, line 71: m, n or n, m?

src/datajudge/constraints/stats.py
```diff
@@ -14,44 +16,94 @@ def __init__(
...
+
+        # calculate approximate p-value
+        p_value = KolmogorovSmirnov2Sample.approximate_p_value(d_statistic, m, n)
+
+        return d_statistic, p_value, n, m
+
+    def test(self, engine: sa.engine.Engine) -> TestResult:
+
+        # get query selections and column names for target columns
+        selection1 = str(self.ref.data_source.get_clause(engine))
```
jonashaag, line 83: Do we want to take the `str` here or rather in `get_ks_2sample`?

src/datajudge/db_access.py
```diff
@@ -902,3 +902,85 @@ def get_column_array_agg(
...
+    """
+    Runs the query for the two-sample Kolmogorov-Smirnov test and returns the test statistic d.
+    """
+    table1_selection, col1 = table1
+    table2_selection, col2 = table2
+
+    if is_mssql(engine):
+        table1_selection = str(table1_selection).replace(
+            '"', ""
+        )  # tempdb.dbo.int_table
```
jonashaag, line 917: Comments are different, accident?

src/datajudge/constraints/stats.py
YYYasin19: 👍

src/datajudge/constraints/stats.py
YYYasin19: You are correct, I think.

src/datajudge/constraints/stats.py
YYYasin19: It actually doesn't matter, but I have corrected the order and renamed the variables to be more concise everywhere.

src/datajudge/db_access.py
YYYasin19: Intention was to show what the replacement should achieve, is now somewhat cleaner.

ivergara:
> > Usually one uses the p value of 0.05 or if you want to be very picky 0.01. A difference in the calculated value via SQL and scipy like you find will not make anyone accept or reject the null hypothesis.
> 
> The problem is that in practice for one of our tests the reported p-value in the magnitude of `1e-53` because the table has 100+ million rows. Therefore, my idea to use scipy in very scientific environments where accuracy is required.

Mh, that's a problem... and it's a rather generalized issue with p-values in general.

Although I did try to look for possible solutions, I didn't find any.

One idea would be to base the data test around the value `D`, the maximum distance between the samples. Then the test receives a threshold for `D` that needs to be fulfilled. Now, I agree that this solution is not very satisfactory as it just kicks the problem to define a sensible threshold and at least I personally have no idea how to interpret `D` enough to be able to choose a number for the threshold.

YYYasin19:
> One idea would be to base the data test around the value `D`, the maximum distance between the samples. Then the test receives a threshold for `D` that needs to be fulfilled. Now, I agree that this solution is not very satisfactory as it just kicks the problem to define a sensible threshold and at least I personally have no idea how to interpret `D` enough to be able to choose a number for the threshold.

My experience is only anectodal as well, but from what I've seen for various sample sizes (`10`, `10_000` and `100_000_000`) the test statistic `d` doesn't get too small. Therfore, I chose `1e-10` as allowed deviation.

src/datajudge/constraints/stats.py
```diff
@@ -72,24 +65,26 @@ def c(alpha: float):
         )
 
     @staticmethod
-    def calculate_statistic(engine, table1, table2) -> Any:
+    def calculate_statistic(engine, table1_def: tuple, table2_def: tuple) -> Any:
```
ivergara, line 68: too vague and weak annotations

YYYasin19: What do you think about 
```py
def calculate_statistic(
        engine,
        table1_def: tuple[Selectable | str, str],
        table2_def: tuple[Selectable | str, str],
    ) -> tuple[float, Optional[float], int, int]:
```

ivergara: Better. I guess `Selectable` is a `sqlalchemy` type. 

Remember that we're still supporting some older versions of python, thus the `|` should be `Union[Selectable, str]`

ivergara: > > Better. I guess Selectable is a sqlalchemy type.
> 
> Do you know what else should be used? In practice, everything with a sensible **str** works

No, for now, it looks good.

jonashaag:
Great work!

PR #33, opened 2022-07-25
Improve varchar regex assertion message

kklein:
Changes:
* Adapt the formatting of the assertion message.
* Cap the number of counterexamples displayed in the assertion message.

Previously:

```AssertionError: string_table's columns  name breaks regex ^QuantCo for example values {'Google', 'BMW', 'Apple'}for a fraction '0.5' over the accepted tolerance '0.0'```

Now:

```AssertionError: string_table's column(s) 'name' breaks regex '^QuantCo' in 0.5 > 0.0 of the cases. In absolute terms, 3 of the 6 samples violated the regex. Some counterexamples consist of the following: ['Apple', 'Google', 'BMW'].```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Changelog
 =========
 
+1.2.0 - 2022.xx.xx
+------------------
+
+**New features**
+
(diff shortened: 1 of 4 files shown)


kklein:
FYI @ivergara

PR #34, opened 2022-07-25
Fix msodbc installation problem in CI.

kklein:
```The following packages will be DOWNGRADED:
msodbcsql17
0 upgraded, 0 newly installed, 1 downgraded, 0 to remove and 13 not upgraded.
E: Packages were downgraded and -y was used without --allow-downgrades.
Error: Process completed with exit code 100.
```

See e.g. here [0] for current problem in CI.

[0] https://github.com/Quantco/datajudge/runs/7494860116?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,6 +79,12 @@ jobs:
           use-mamba: true
           environment-file: environment.yml
           activate-environment: datajudge
+      # TODO: Consider removing this manual installation since the docker image already comes with a msodbc driver.
+      - name: Install msodbcsql17 driver
+        shell: bash -l {0}
+        run: |
+          wget https://packages.microsoft.com/ubuntu/20.04/prod/pool/main/m/msodbcsql17/msodbcsql17_17.9.1.1-1_amd64.deb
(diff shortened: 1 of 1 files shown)


jonashaag:
Interesting. If we already have a version of `msodbcsql17` installed, why even install another version?

kklein:
My interpretation was that some other dependency would be downgraded, not `msodbcsql17` itself - I'm quite unsure about this, though.

jonashaag:
No:

```
The following packages will be DOWNGRADED:
  msodbcsql17
```

Can you try removing the entire `msodbcsql17` install/upgrade/downgrade?

kklein:
I'd like to think that our docker image simply comes with the driver already installed. Yet, I'm not sure why we added this installation step.

I found a dockerfile installing `msodbcsql18`:

https://github.com/microsoft/mssql-docker/blob/master/linux/mssql-tools/Dockerfile.ubuntu2004#L15

I think I would be in favor of sticking with the downgrade and delaying the decision as to whether to get rid of the installation step until @ivergara is back. Does that sound good to you @jonashaag ?

jonashaag:
Sure.

PR #38, opened 2022-07-26
Provide implementation of in-database `VarCharRegex` constraint.

kklein:
The current implementation of `VarCharRegex` fetches all rows from database to memory and matches them against a regex via a Python implementation.

This has several downsides:
- Heavy/slow on network in the case of snowflake
- Substantially larger memory cost
- Usually longer runtime
- Inconsistency with the datajudge paradigm to do heavy lifting in database

Instead, this PR seeks to offer the in-database matching of regexes for some database systems. More precisely, this change relies on `sqlalchemy`'s relatively new `regexp_match`[0] method. According to their documentation:

> Regular expression support is currently implemented for Oracle, PostgreSQL, MySQL and MariaDB. Partial support is available for SQLite. Support among third-party dialects may vary.

As of now, tests for the in-database computation run only against Postgres. I also ran them locally against Snowflake, which succeeded but relies on `snowflake-sqlalchemy` version 1.4.0. As long as this [1] bug isn't fixed, we can't enable the tests against snowflake in CI since we still need version <1.4.0.

Put differently, an eager user can install `snowflake-sqlalchemy` version 1.4.0 and make use of this feature.

Screenshot 2022-07-27 at 12 49 38 PM

[0] https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.regexp_match
[1] https://github.com/Quantco/datajudge/issues/39

### DIFF
--- a/src/datajudge/constraints/varchar.py
+++ b/src/datajudge/constraints/varchar.py
@@ -34,12 +34,12 @@ class VarCharRegexDb(Constraint):
             regex=self.regex,
             n_counterexamples=self.n_counterexamples,
         )
-        if not self.aggregated:
-            n_rows, n_rows_selections = db_access.get_row_count(engine=engine, ref=ref)
-        else:
+        if self.aggregated:
             n_rows, n_rows_selections = db_access.get_unique_count(
                 engine=engine, ref=ref
             )
+        else:
+            n_rows, n_rows_selections = db_access.get_row_count(engine=engine, ref=ref)
         return (
             n_violations,
             n_rows,
--- a/src/datajudge/constraints/varchar.py
+++ b/src/datajudge/constraints/varchar.py
@@ -9,6 +9,62 @@ from ..db_access import DataReference
 from .base import Constraint, OptionalSelections, TestResult
 
 
+class VarCharRegexDb(Constraint):
+    def __init__(
+        self,
+        ref: DataReference,
+        regex: str,
+        relative_tolerance: float = 0.0,
+        aggregated: bool = True,
+        n_counterexamples: int = 5,
+    ):
+        super().__init__(ref, ref_value=relative_tolerance)
+        self.regex = regex
+        self.aggregated = aggregated
+        self.n_counterexamples = n_counterexamples
+
+    def retrieve(self, engine: sa.engine.Engine, ref: DataReference):
+        (
+            n_violations,
+            counterexamples,
+        ), violations_selections = db_access.get_regex_violations(
+            engine=engine,
+            ref=ref,
+            aggregated=self.aggregated,
+            regex=self.regex,
+            n_counterexamples=self.n_counterexamples,
+        )
+        if not self.aggregated:
+            n_rows, n_rows_selections = db_access.get_row_count(engine=engine, ref=ref)
+        else:
+            n_rows, n_rows_selections = db_access.get_unique_count(
+                engine=engine, ref=ref
+            )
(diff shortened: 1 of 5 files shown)


src/datajudge/requirements.py
```diff
@@ -606,18 +608,42 @@ def add_varchar_regex_constraint(
...
-                regex,
-                allow_none=allow_none,
-                relative_tolerance=relative_tolerance,
-                aggregated=aggregated,
-                n_counterexamples=n_counterexamples,
+        if computation_in_db:
+            if allow_none:
+                warnings.warn(
+                    """
+                    Parameter 'allow_none' is not allowed when using 'computation_in_db'.
```
kklein, line 622: From my perspective it would probably make sense to deprecate this parameter both when using `computation_in_db` and when not using it.

src/datajudge/constraints/varchar.py
```diff
@@ -9,6 +9,69 @@
...
+            n_violations,
+            counterexamples,
+        ), violations_selections = db_access.get_regex_violations(
+            engine=engine,
+            ref=ref,
+            aggregated=self.aggregated,
+            regex=self.regex,
+            n_counterexamples=self.n_counterexamples,
+        )
+        if not self.aggregated:
```
jonashaag, line 37: Suggest to change the cases:

```
if not ...:
else:

->

if:
else:
```

src/datajudge/db_access.py
```diff
@@ -988,3 +988,42 @@ def get_ks_2sample(
...
+        counterexamples_selection = violation_selection.limit(n_counterexamples)
+    else:
+        raise ValueError(f"Unexpected number of counterexamples: {n_counterexamples}")
+
+    if counterexamples_selection is not None:
+        selections.append(counterexamples_selection)
+
+    with engine.connect() as connection:
+        n_violations_result = connection.execute(n_violations_selection).scalar()
+        if counterexamples_selection is not None:
```
jonashaag, line 1022: See above.

src/datajudge/requirements.py
```diff
@@ -592,6 +593,7 @@ def add_varchar_regex_constraint(
         relative_tolerance: float = 0.0,
         aggregated: bool = True,
         n_counterexamples: int = 5,
+        computation_in_db=False,
```
jonashaag, line 596: I suggest to make this two separate `add_...` functions instead. Maybe call one "db" and the other "py" and deprecate the current name.

jonashaag (approved):
Looks great.

src/datajudge/requirements.py
kklein: Unfortunately we cannot deprecate the in-memory version since there is no sqlalchemy implementation to do regex matching for mssql.

Aside from that I also wondered whether to switch with different methods or via parameters and was unsure. Not trying to oppose you suggestion: How do you see that trade-off?

jonashaag: Hard to come up with a general rule. Indicators that something should be a different function/method/program:

- Different interface: Python-style regexes vs. database specific regexes
- Behavior is substantially different: Moved work (= CPU and memory resources used) from Python to database.
- Parameter combinations that are not applicable: `allow_none` is not applicable to the database version.

kklein:
FYI @ivergara

kklein:
Thanks a lot @jonashaag :)

PR #43, opened 2022-07-28
Minor improvements on readthedocs

kklein:
Before:
Screenshot 2022-07-28 at 6 16 11 PM

After:
Screenshot 2022-07-28 at 6 16 32 PM

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -8,7 +8,13 @@ author = "QuantCo Inc."
 release = "1.0"
 version = "1.0.0"
 
-extensions = ["numpydoc", "sphinxcontrib.apidoc"]
+extensions = [
+    "numpydoc",
+    "sphinxcontrib.apidoc",
+    "sphinx.ext.autodoc",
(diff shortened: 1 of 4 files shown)


ivergara:
Why would you say it's an improvement to not have the types in the signature but in a separate block instead?

kklein:
@ivergara I acknowledge that in principle it is difficult to make a priori statements about superiority in terms of UI matters without user tests.

The upside I personally saw with the `sphinx_autodoc_typehints` change was that I find it easier to parse what the parameters of a function/method/class are if they are displayed as a list as compared to a sequence of words.

Afaict, this list representation is often achieved by having a dedicated parameters block as part of the doc string, e.g. [0]. 

My thought was that we could gain the advantage of having a similar representation without having to rewrite all of our docstrings by using this very sphinx extension.

[0] https://github.com/Quantco/glum/blob/main/src/glum/_glm.py#L126-L132

ivergara (approved):

Issue #42, opened 2022-07-28
`KolmogorovSmirnov2Sample`: `Condition`s are ignored


kklein:
The `KolmogorovSmirnov2Sample` constraint can be initialized with `DataReference`s containing `Condition`s. This is apparent by the parameters in `add_ks_2sample_constraint`:

https://github.com/Quantco/datajudge/blob/main/src/datajudge/requirements.py#L1276-L1277

Yet, since `get_ks_2sample` in `db_access` operates on `DataSource` objects - instead of `DataReference` objects - the `Condition`s attached to `DataReference`s are ignored.

While a `Condition` was indeed used in one integration test, it was used for both `DataReference`s at the same time. Since the same test also succeeds without a `Condition` it did not became apparent that `Condition`s are ignored:

https://github.com/Quantco/datajudge/blob/main/tests/integration/test_integration.py#L1768-L1789

Resolving this issue is one of the many things we should obtain for free when tackling https://github.com/Quantco/datajudge/issues/29 and only ever accessing the underlying data via `DataReference.get_selection` - which takes care of these things under the hood.

PR #44, opened 2022-07-28
Reimplement Kolmogorov Smirnov query logic with sqlalchemy's Language Expression API

kklein:
See linked issues for context.

Importantly, I have high hopes that the isolation and explicit testing of `_cross_cdf_selection` will now be useful for other kinds of `Constraint`s, e.g. https://github.com/Quantco/datajudge/issues/45.

### DIFF
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,6 +1,6 @@
 import math
 import warnings
-from typing import Optional, Tuple
+from typing import Optional
 
 import sqlalchemy as sa
 
@@ -73,50 +73,63 @@ class KolmogorovSmirnov2Sample(Constraint):
         engine,
         ref1: DataReference,
         ref2: DataReference,
-    ) -> Tuple[float, Optional[float], int, int]:
+    ) -> tuple[float, Optional[float], int, int, list]:
 
         # retrieve test statistic d, as well as sample sizes m and n
-        d_statistic = db_access.get_ks_2sample(
+        d_statistic, ks_selections = db_access.get_ks_2sample(
             engine,
             ref1,
             ref2,
         )
 
-        n_samples, _ = db_access.get_row_count(engine, ref1)
-        m_samples, _ = db_access.get_row_count(engine, ref2)
+        n_samples, n_selections = db_access.get_row_count(engine, ref1)
+        m_samples, m_selections = db_access.get_row_count(engine, ref2)
 
         # calculate approximate p-value
         p_value = KolmogorovSmirnov2Sample.approximate_p_value(
             d_statistic, n_samples, m_samples
         )
 
-        return d_statistic, p_value, n_samples, m_samples
+        selections = n_selections + m_selections + ks_selections
+        return d_statistic, p_value, n_samples, m_samples, selections
 
     def test(self, engine: sa.engine.Engine) -> TestResult:
-
-        # get query selections and column names for target columns
-
-        d_statistic, p_value, n_samples, m_samples = self.calculate_statistic(
+        (
+            d_statistic,
+            p_value,
+            n_samples,
+            m_samples,
+            selections,
+        ) = self.calculate_statistic(
             engine,
             self.ref,
             self.ref2,
         )
-
-        # calculate test acceptance
         result = self.check_acceptance(
             d_statistic, n_samples, m_samples, self.significance_level
         )
 
         assertion_text = (
             f"Null hypothesis (H0) for the 2-sample Kolmogorov-Smirnov test was rejected, i.e., "
-            f"the two samples ({self.ref.get_string()} and {self.target_prefix})"
-            f" do not originate from the same distribution."
+            f"the two samples ({self.ref.get_string()} and {self.target_prefix}) "
+            f"do not originate from the same distribution. "
             f"The test results are d={d_statistic}"
         )
         if p_value is not None:
-            assertion_text += f"and {p_value=}"
+            assertion_text += f" and {p_value=}"
+        assertion_text += "."
+
+        if selections:
+            queries = [
+                str(selection.compile(engine, compile_kwargs={"literal_binds": True}))
+                for selection in selections
+            ]
 
         if not result:
-            return TestResult.failure(assertion_text)
+            return TestResult.failure(
+                assertion_text,
+                self.get_description(),
+                queries,
+            )
 
         return TestResult.success()
--- a/src/datajudge/constraints/stats.py
+++ b/src/datajudge/constraints/stats.py
@@ -1,9 +1,8 @@
 import math
 import warnings
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple
 
 import sqlalchemy as sa
-from sqlalchemy.sql import Selectable
 
 from .. import db_access
 from ..db_access import DataReference
@@ -71,13 +70,15 @@ class KolmogorovSmirnov2Sample(Constraint):
     @staticmethod
     def calculate_statistic(
         engine,
-        table1_def: Tuple[Union[Selectable, str], str],
-        table2_def: Tuple[Union[Selectable, str], str],
+        ref1: DataReference,
+        ref2: DataReference,
     ) -> Tuple[float, Optional[float], int, int]:
 
         # retrieve test statistic d, as well as sample sizes m and n
         d_statistic, n_samples, m_samples = db_access.get_ks_2sample(
-            engine, table1=table1_def, table2=table2_def
+            engine,
+            ref1,
+            ref2,
         )
 
         # calculate approximate p-value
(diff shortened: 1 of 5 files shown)


src/datajudge/constraints/stats.py
```diff
@@ -63,9 +63,10 @@ def check_acceptance(
         def c(alpha: float):
             return math.sqrt(-math.log(alpha / 2.0 + 1e-10) * 0.5)
 
-        return d_statistic <= c(accepted_level) * math.sqrt(
+        threshold = c(accepted_level) * math.sqrt(
```
kklein, line 66: Slightly more convenient to debug.

tests/integration/test_integration.py
```diff
@@ -1925,31 +1913,41 @@ def test_ks_2sample_constraint_perfect_between(engine, int_table1, data):
     assert operation(test_result.outcome), test_result.failure_message
 
 
-# TODO: Enable this test once the bug is fixed.
-@pytest.mark.skip(reason="This is a known bug and unintended behaviour.")
```
kklein, line 1929: No longer skipping this test (as well as adding further examples) should indicate that this PR solves https://github.com/Quantco/datajudge/issues/42

src/datajudge/db_access.py
```diff
@@ -904,77 +904,149 @@ def get_column_array_agg(
     return result, selections
 
 
+def _cdf_selection(engine, ref: DataReference, cdf_label: str, value_label: str):
+    col = ref.get_column(engine)
+    selection = ref.get_selection(engine).subquery()
+
+    # Step 1: Calculate the CDF over the value column.
```
jonashaag, line 911: Just curious: Is possible to merge the two steps? Like so

```py
sa.select([
  selection.c[col],
  sa.func.max(sa.func.cume_dist().over(order_by=col)),
])
.group_by(selection.c[col])
```

src/datajudge/constraints/stats.py
```diff
@@ -63,59 +63,73 @@ def check_acceptance(
...
         )
+        return d_statistic <= threshold
 
     @staticmethod
     def calculate_statistic(
         engine,
         ref1: DataReference,
         ref2: DataReference,
-    ) -> Tuple[float, Optional[float], int, int]:
+    ) -> Tuple[float, Optional[float], int, int, List]:
```
jonashaag, line 76: List of what?

src/datajudge/db_access.py
```diff
@@ -904,77 +904,149 @@ def get_column_array_agg(
...
+    )
+    return filled_cross_cdf, cdf_label1, cdf_label2
+
+
 def get_ks_2sample(
     engine: sa.engine.Engine,
     ref1: DataReference,
     ref2: DataReference,
-) -> float:
+):
```
jonashaag, line 1027: Do you not annotate return types?

jonashaag (commented):
I find the SQLAlchemy code incredibly difficult to follow. Not saying that it is because you are writing complicated SQLAlchemy. Maybe it's just more difficult for me (with little SQLAlchemy experience) to follow than to follow SQL. Maybe it would help people like me if we had a simplified version of the query in SQL somewhere in a docstring to help get an overview of the code.

src/datajudge/db_access.py
kklein: I wondered the same and doing it that way leads to an error:
`sqlalchemy.exc.ProgrammingError: (psycopg2.errors.GroupingError) aggregate function calls cannot contain window function calls`

src/datajudge/constraints/stats.py
kklein: Quite frankly we/I haven't figured out yet what the latest common SQLAlchemy ancestor type yet. Throughout almost all of `db_access` we don't annotate the type of the selections because the types, iirc, differ. Quite a few are simply `sqlalchemy.sql.select.Selectable`. Yet, some aren't and still support the necessary method interfaces.

Now there certainly are remedies to this situation but we haven't considered this to be 'sufficiently important' up until now.

src/datajudge/db_access.py
kklein: See above.

kklein:
> I find the SQLAlchemy code incredibly difficult to follow. Not saying that it is because you are writing complicated SQLAlchemy. Maybe it's just more difficult for me (with little SQLAlchemy experience) to follow than to follow SQL. 

I definitely understand where you're coming from. A couple of hopefully related comments from my side:
* Personally, I see multiple upsides of using the language expression API. 
  * Firstly, it conveniently enables the use of abstractions such as `ref.get_selection(engine)`. IIrc not being able to use this abstraction caused a lot of the development effort in the raw-sql version of this query logic. Moreover, it caused a bug in relation to `Condition`s. 
  * Secondly, some raw sql queries might actually run just fine against multiple dialects. Yet, this is definitely not the case for all of our queries. When it is not possible, the language expression API actually allows us to 'write once, run everywhere™'. I'm not sure whether having a patchwork of raw sql and language expression api would be desirable in the long run. 
  * Thirdly, I find it easier to share logic between different queries than with raw sql queries - see comment about reusing `_cross_cdf_selection` for another hypothesis test.

* As you gently allude to, I think I have indeed seen a development on my end when it comes to reading language expression code. I believe to find it easier to read now than two years ago.
* I still don't find it easy to read.
* To me, this very example seems to be on one side of the query logic complexity spectrum, at least in how far datajudge is concerned.
* Personally - even though I admire the rigid structure and invaluable comments introduced by @YYYasin19 - I didn't find the raw sql code easy to follow either. It took me some debugging and reimplementing to understand some of the intricacies. 

> Maybe it would help people like me if we had a simplified version of the query in SQL somewhere in a docstring to help get an overview of the code.

I'm definitely open to the idea. 

I made a point out of mostly preserving @YYYasin19's structure of iterative steps - in contrast to recursion - to allow the curious developer to create, read and run these intermediate steps. What I often do is to set breakpoints and run `str(intermediate_selection)` or even `engine.connect().execute(intermediate_selection)` via `pdb` while executing the integration tests. Moreover, I hoped to have made the query building a little more clear with this[0] example as well as the dedicated test case[1].

Nevertheless, if you have a concrete suggestion I'm happy to follow your lead. :)

[0] https://github.com/Quantco/datajudge/pull/44/files#diff-74c4ccf9b9cba732c4562df6e1e05a5ecdee4e9c04f38276b54e85f850155660R943-R945
[1] https://github.com/Quantco/datajudge/pull/44/files#diff-294af7c1d98bcac87ae111bebc4c13a1b1d8155cce13ea48446397e950f9db32R7-R28

src/datajudge/db_access.py
ivergara: Could you add in the docstring a bit more information about the objective/idea behind this method? 

It's great to have the comments on the step-by-step like in the SQL version before, but a summary would be a great addition to it, particularly clarifying and being explicit about the meaning of the arguments.

src/datajudge/db_access.py
```diff
@@ -904,77 +904,149 @@ def get_column_array_agg(
...
+
+    return grouped_cdf_selection
+
+
+def _cross_cdf_selection(
+    engine, ref1: DataReference, ref2: DataReference, cdf_label: str, value_label: str
+):
+    """Create a cross cumulative distribution function selection given two samples.
+
+    Concretely, both ``DataReference``s are expected to have specified a single relevant column.
```
ivergara, line 939: Don't you want to explicitly enforce that expectation at the beginning of the method?

ivergara (approved):
Great job!

ivergara:
> > Maybe it would help people like me if we had a simplified version of the query in SQL somewhere in a docstring to help get an overview of the code.
> 
> I'm definitely open to the idea.

Why not put a link to the full version in Yasin's repository containing the SQL version?

YYYasin19:
Very nice (re-)implementation in the expression language API!
I agree with the fact that it is hard to parse without having experience with the API. I think we can split the query up as much as possible here, since the final compilation and optimization will still be done by the database engine(s), so our goal should be readability and maintainability -- though I don't see any obvious wins left here.

src/datajudge/db_access.py
YYYasin19: Am I missing a comment somewhere? What's the idea? Shouldn't we annotate as much as possible?
Even using Any makes sense because you are proactively declaring that you don't care while not annotating leaves the user to guess where it's (1) unknown (2) not important (3) missing

YYYasin19 (approved):

src/datajudge/db_access.py
kklein: > Am I missing a comment somewhere?
Have you read this [0]?

>  because you are proactively declaring that you don't care 

It's not clear to me why we don't care.

[0] https://github.com/Quantco/datajudge/pull/44#discussion_r933778546

src/datajudge/db_access.py
kklein: Good idea.
[f396053](https://github.com/Quantco/datajudge/pull/44/commits/f3960539612780cfa26aea64e508d6e48bde9b32)

src/datajudge/db_access.py
kklein: Yeah, great point! Did it for all at once:
[3c9408c](https://github.com/Quantco/datajudge/pull/44/commits/3c9408c54255a83cf539134f4b5e2086e45f6cad)

src/datajudge/db_access.py
kklein: I'll take the liberty to merge for now. Yet, if you consider this an open topic still, happy to further discuss this and address it as a follow-up @YYYasin19 .

src/datajudge/db_access.py
```diff
@@ -288,7 +288,13 @@ def get_column(self, engine):
                 f"Trying to access column of DataReference "
                 f"{self.get_string()} yet none is given."
             )
-        return self.get_columns(engine)[0]
+        columns = self.get_columns(engine)
```
jonashaag, line 291: I like this one

```py
(col,) = self.get_columns(engine)
return col
```

PR #52, opened 2022-08-15
Restrict writing of test data to scenario in which it previously didn't exist.

kklein:
See linked issue.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -52,8 +52,9 @@ def metadata():
 
 
 def _handle_table(engine, metadata, table_name, columns, data):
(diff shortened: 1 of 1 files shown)


ivergara (approved):
I'm not opposed to this change, but what's the advantage you see in this?

In case this gets merged, we'd have to be aware that changes in the data of existing tables will require manual intervention in Snowflake.

kklein:
I would've believed this change to correspond to what you said:

> I do like option 3

> Now, my question is, do we really need/want teardown code?

Did I misunderstand?

ivergara:
> > Now, my question is, do we really need/want teardown code?
> 
> Did I misunderstand?

No. Although now I remember the origin, multiple runs in Snowflake screwing each other. This should fix it. 

Merge away!

PR #55, opened 2022-08-26
Remove snowflake-sqlalchemy pin.

kklein:
See https://github.com/snowflakedb/snowflake-sqlalchemy/issues/325

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,9 +18,7 @@ dependencies:
   - sqlalchemy
   - psycopg2
   - freetds
-  # Pinned at least until this bug is resolved:
-  # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/325
-  - snowflake-sqlalchemy=1.3.4
+  - snowflake-sqlalchemy
   - flit-core
   - flit
   - pytest-cov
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -18,9 +18,7 @@ dependencies:
   - sqlalchemy
   - psycopg2
   - freetds
-  # Pinned at least until this bug is resolved:
-  # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/325
-  - snowflake-sqlalchemy=1.3.4
+  - snowflake-sqlalchemy
```
ivergara, line 21: Wouldn't maybe make sense to exclude the "broken" version here?

ivergara (approved):

environment.yml
kklein: I thought about that too. I then figured that technically speaking, there are likely many versions of various dependencies of ours that would break the code. We don't maintain a ruleset or blacklist for other dependencies either.
At the end of the day this is the development environment and everything should work if a developer updates their environment. I figured that'd be good enough.
Do you disagree?

kklein: This is all assuming that one doesn't go down the road of the 'all dependencies should always be pinned' philosophy - which is certainly a valid point of discussion but not something we have touched upon up until now with datajudge.

ivergara: Fair enough. We don't have to have that discussion now. Merge away then!

PR #56, opened 2022-09-10
Expand on testing in documentation

kklein:
Documentation build for this PR can be seen here:
https://datajudge--56.org.readthedocs.build/en/56/testing.html

Datajudge delegates the execution of tests to pytest. As a consequence one might argue that there is no need to document the execution of tests since the user might want to familiarize themselves with pytest independently of the datajudge documentation.

Yet, in practice, it might be desirable to give current and potential users ideas and guidance with respect to configuring the test execution adaptation.

So far, our documentation carries little information on how to configure test execution. This PR introduces an entire documentation article with a couple of exemplary adaptations. These adaptations have been relevant for some of our internal users.


docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+one failing test to halt all others.
+
+Many of these approaches rely on adapting pytest's ``conftest.py``. If you are not familiar
+with this concept, you might want to read up on it
+`here <https://docs.pytest.org/en/6.2.x/writing_plugins.html#conftest-py-plugins>`_.
+
+Subselection
+------------
+
+Most often one might want to run all tests implied by a specification.
```
ivergara, line 29:
```suggestion
Most often one might want to run all tests defined by a specification.
```

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Yet, for example after believing to have fixed a data problem, one might simply
+want to test whether a single test, which had previously been failing, succeeds
+at last.
+
+Another example for when one would like to test a subset of tests is if the data at
+hand is not available in its entirety. Rather, it could be that one would like
+to run a subset of the test suite against a subsample of the typical dataset.
+
+In this section we present to approaches to do subselection of tests.
```
ivergara, line 39:
```suggestion
In this section, we present two approaches to do a subselection of tests.
```

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+   markers = basic: basic specification
+	     all: entire specification
+
+Once that is taken care of, one can adapt one's specification as follows:
+
+.. code-block:: python
+  :caption: ``specification.py``
+
+  def get_basic_requirements() -> List[Requirement]:
+      ...
```
ivergara, line 112: Is this `...` standing for "add your requirement and constraints"? 

If so, I'd be more explicit since `...` is actually a valid Python syntax.

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+
+  def pytest_addoption(parser):
+      parser.addoption("--new_db", action="store", help="name of the new database")
+      parser.addoption("--old_db", action="store", help="name of the old database")
+
+
+  def pytest_generate_tests(metafunc):
+      params = {
+          "db_name_new": metafunc.config.option.new_db,
+          "db_ne_old": metafunc.config.option.old_db,
```
ivergara, line 181:
```suggestion
          "db_name_old": metafunc.config.option.old_db,
```
?

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+
+  $ pytest specification.py -m all
+
+to test all ``Requirement`` s.
+
+
+Parametrization
+---------------
+
+A given specification might rely on identifiers such as database names or
```
ivergara, line 152: I find the two first paragraphs a bit too abstract. And while the first one might be fine as an opening one, the second doesn't help much in supporting and making it more concrete.

Once the example comes, it's more understandable, but I also am familiar with the intention.

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+
+.. code-block:: python
+  :caption: ``conftest.py``
+
+  def pytest_generate_tests(metafunc):
+      if "basic_constraint" in metafunc.fixturenames:
+           metafunc.parametrize(
+	       "basic_constraint",
+               metafunc.module.get_basic_constraints(),
+               ids=metafunc.module.idfn,
```
ivergara, line 83: You're invoking this `idfn` callable, but in this local context, the one of this example, it's never defined.

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+      return requirements
+
+
+  def get_constraints(params):
+      return [
+	  constraint for requirement in get_requirements(params) for constraint in requirement
+      ]
+
+
+  def idfn(constraint):
```
ivergara, line 211: Now, in this example you define it, and we see it can be a very simple function. But its purpose is not explained anywhere, one would have to find that in the `pytest_generate_tests` is being used and then go and look in `pytest`s documentation what does it mean.

Maybe a bit more explanation between prose in the documentation and in code comment might help the reader.

docs/source/testing.rst
```diff
@@ -0,0 +1,316 @@
...
+By default, running ``pytest`` tests will output test results to one's respective shell.
+Alternatively, one might want to generate an html report summarizing and expanding on
+all test results. This can be advantageous for
+
+* Sharing test results with colleagues
+* Archiving and tracking test results over time
+* Make underlying sql queries conveniently accessible
+
+Concretely, such an html report can be generated by
+`pytest-html <https://github.com/pytest-dev/pytest-html>`_. Once installed, using it as simple
```
ivergara, line 238:
```suggestion
`pytest-html <https://github.com/pytest-dev/pytest-html>`_. Once installed, using it is as simple
```

ivergara (approved):
Great idea!

YYYasin19 (approved):
Very nice! 🎉 
I have nothing to add to this text specifically outside the things @ivergara already mentioned.

docs/source/testing.rst
YYYasin19: Though I easily understood them on the second read, I'd agree.
Maybe something more like 

---
A given _test_ specification might rely on identifiers such as database names or table names. Moreover it might be that, e.g. when iterating from one _version_ of the data to another, these _names_ change. ~All the while the expectations towards how the data relates to another might remain.~

In other words, it could be that the test logic ~logic of the specification~ should remain unchanged while its pointers to data might change. Therefore, one ~One might simply rely on hard-coded references to data. Yet, one~ might just as well parametrize such pointers or identifiers.

---
Or a much shorter version might also suffice:

> For applying the same test suite to different data sources and targets, e.g. different versions over time, one might want to parameterize pointers such as database names or table identifiers.

ivergara: Somethign I forgot to mention is that `parametrization` here clashes with what a `parametrize` mark in `pytest`.

I think you can safely use the word `parameter`, but not `parametrization` which might cause some confusion at first. To me, what it's happening here evokes the concept of `templating` in a way.

docs/source/testing.rst
kklein: Great point!

docs/source/testing.rst
kklein: Nice catch :)

docs/source/testing.rst
kklein: Thanks a lot for your thoughts, both of you. I tried to distill them into an altered version of this section.

kklein:
Thanks both of you!

PR #57, opened 2022-09-16
Add example on dates data

kklein:
Sadly, sqlite doesn't offer date data types - otherwise I would've added coded to actually run this, as we have in the companies example[0].

Eventually, one might find a way around this, e.g. by offering a script to upload data to a postgres database - as we do in the twitch example[1].

I figured this PR would be a good start. If someone feels like turning this into an executable specification in a follow-up PR, this would of course be highly appreciated.

Rendered version: https://datajudge--57.org.readthedocs.build/en/57/examples/example_dates.html

[0] https://datajudge.readthedocs.io/en/latest/examples/example.html
[1] https://datajudge.readthedocs.io/en/latest/examples/example_twitch.html

### DIFF
--- a/docs/source/examples/example_dates.rst
+++ b/docs/source/examples/example_dates.rst
@@ -1,10 +1,16 @@
 Example: Dates
 ==============
 
-In this example we will assume a table containing prices for a given product.
-Importantly, these prices are valid for a certain date range only. Here we
-assume that the price is indicated in the ``price`` column, the date from which it
-is valid - the date itself included - in ``date_from`` and the the until when
+This example concerns itself with expressing constraints against data revolving
+around dates. While constraints between table with respect to dates exist,
+we will only illustrate constraints on a single table and reference values. As
+a consequence, we will only use ``WithinRequirement``, in contrast to
+``BetweenRequirement``.
+
+Concretely, we will assume a table containing prices for a given product.
+Importantly, these prices are valid for a certain date range only. More precisely,
+we assume that the price is indicated in the ``price`` column, the date from which
+it is valid - the date itself included - in ``date_from`` and the the until when
 it is valid - the date itself included - in the ``date_to`` column.
 
 Such a table might look as follows:
@@ -138,4 +144,15 @@ Assuming that such a table exists in database, we can write a specification agai
     # up these tests.
     test_constraint = collect_data_tests(requirements)
 
+Pleas note that the ``DateNoOverlap`` and ``DateNoGap`` constraints also exist
+in a slightly different form: ``DateNoOverlap2d`` and ``DateNoGap2d``.
+As the names suggest, these can operate in 'two date dimensions'.
+It could for instance be, that a table has four date columns, representing two
+ranges in distinct dimensions, respectively:
+* ``date_from``: Date from when a price is valid
+* ``date_to``: Date until when a price is valid
+* ``date_definition_from``: Date when a price definition was inserted
+* ``date_definition_to``: Date until when a price definition was used
 
+Analogously to the unidimensional scenario illustrated here, one might care
+for certain constraints in two dimensions.
--- a/docs/source/examples/example_dates.rst
+++ b/docs/source/examples/example_dates.rst
@@ -1,11 +1,10 @@
 Example: Dates
 ==============
 
-This example concerns itself with expressing constraints against data revolving
-around dates. While constraints between table with respect to dates exist,
-we will only illustrate constraints on a single table and reference values. As
-a consequence, we will only use ``WithinRequirement``, in contrast to
-``BetweenRequirement``.
+This example concerns itself with expressing ``Constraint`` s against data revolving
+around dates. While date ``Constraint`` s between tables exist, we will only illustrate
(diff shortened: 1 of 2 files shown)


docs/source/examples/example_dates.rst
```diff
@@ -0,0 +1,158 @@
+Example: Dates
+==============
+
+This example concerns itself with expressing ``Constraint`` s against data revolving
```
ivergara, line 4: Why do you tend to have a space character when making plurals of "formatted" words? Like here in `` Constraint`` s?

I saw the rendered version, and I think I can see a space between the word and the last letter.

docs/source/examples/example_dates.rst
```diff
@@ -0,0 +1,158 @@
...
+around dates. While date ``Constraint`` s between tables exist, we will only illustrate
+``Constraint`` s on a single table and reference values here. As a consequence, we will
+only use ``WithinRequirement``, as opposed to ``BetweenRequirement``.
+
+Concretely, we will assume a table containing prices for a given product.
```
ivergara, line 9: To make it more realistic, I'd say it's a products table and add a product id to each row. In this example, all the ids will be the same, but it's clearer even if it's not adding any information.

docs/source/examples/example_dates.rst
```diff
@@ -0,0 +1,158 @@
...
+   * - 13.37
+     - 22/01/16
+     - 22/01/31
+
+Given this table, we would like to ensure that 6 constraints are satisfied:
+
+1. All values from column ``date_from`` should be in January 2022.
+2. All values from column ``date_to`` should be in January 2022.
+3. The minimum value in column ``date_from`` should be the first of January 2022.
+4. The maximum value in column ``date_to`` should be the 31st of January 2022.
```
ivergara, line 38: Maybe you want to explain why you want to have those? Particularly, 1(2) and 3(4) are basically the same. Is there any advantage to doing it like that or just for educational purposes to extract more constraints to show?

docs/source/examples/example_dates.rst
```diff
@@ -0,0 +1,158 @@
...
+5. There is no gap between ``date_from`` and ``date_to``. In other words, every date
+   of January can be assigned to at least one row.
```
ivergara, line 40:
```suggestion
   of January has to be assigned to at least one row.
```

docs/source/examples/example_dates.rst
```diff
@@ -0,0 +1,158 @@
...
+    # we would create another Requirement object.
+    prices_req = WithinRequirement.from_table(
+        db_name="example", schema_name=None, table_name="prices"
+    )
+
+    # Constraint 1:
+    # All values from column date_from should be in January 2022.
+    prices_req.add_date_between_constraint(
+        column="date_from",
+	lower_bound="'20220101'",
```
ivergara, line 67: Not a discussion for now, but cannot help to see this and have thoughts.

This format for a date works, but shouldn't we be supporting ISO format?

Also, it's strictly a date or can this constraint use datetimes?

docs/source/examples/example_dates.rst
```diff
@@ -0,0 +1,158 @@
...
+	use_upper_bound_reference=True,
+    )
+
+    # Constraint 5:
+    # There is no gap between date_from and date_to. In other words, every date
+    # of January can be assigned to at least one row.
+    prices_req.add_date_no_gap_constraint(
+        start_column="date_from",
+	end_column="date_to",
+	key_columns=["price"],
```
ivergara, line 113: Maybe you want to explain here the role of this argument.

ivergara (approved):
Only minor comments.

docs/source/examples/example_dates.rst
kklein: > I saw the rendered version, and I think I can see a space between the word and the last letter.

Same here - I don't like the look of it.

Afaict it is not valid rst to add a character right after a closing double-tick. 

If I do, the build gets very messy:

Screenshot 2022-09-19 at 10 44 07 AM

docs/source/examples/example_dates.rst
kklein: Great idea!

docs/source/examples/example_dates.rst
kklein: > or just for educational purposes to extract more constraints to show

Exactly! Just as the other examples, these are merely illustrations of datajudge functionalities - as opposed to real life case studies.

docs/source/examples/example_dates.rst
kklein: Yeah good point! This was actually a mistake - fixed now.

docs/source/examples/example_dates.rst
kklein: Done

docs/source/examples/example_dates.rst
kklein: I created an issue:
https://github.com/Quantco/datajudge/issues/58

ivergara (approved):

PR #59, opened 2022-09-20
Remove internal interfaces from api doc.

kklein:


docs/source/conf.py
```diff
@@ -19,6 +19,10 @@
 apidoc_module_dir = "../../src/datajudge"
 apidoc_output_dir = "api"
 apidoc_separate_modules = True
+apidoc_excluded_paths = [
+    "../../src/datajudge/db_access.py",
+    "../../src/datajudge/constraints",
```
ivergara, line 24: I'd argue to leave the constraints out of the exclusion.

Although they're woefully un/under-documented, I think that to whoever wants to do something on top of `datajudge` there is where they should go looking. Thus, at least leaving them in the autogenerated documentation for discoverability seems positive to me.

ivergara (approved):

docs/source/conf.py
kklein: Okay, interesting.

> I think that to whoever wants to do something on top of datajudge there is where they should go looking.

I would've thought that such a potential future user would need to work with source code anyway. 
On the other hand, the current users are, afaict, merely interfering with requirements objects.

My main concern is that the api documentation is currently extremely cluttered for our existing users and I wanted to take some of that away.

ivergara: Fair enough.

kklein: Alright, then I'll merge this for now. Yet, I'm very open to reopening the discussion in case you ever have second thoughts. :)

PR #60, opened 2022-09-23
Publish on release

kklein:
One advantage of shifting the deployment to github is that the secrets are collected in a central location, as opposed to locally held by the PyPI maintainers.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -7,8 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: true
-      matrix:
-        PYTHON_VERSION: ['3.8']
+      #matrix:
+      #  PYTHON_VERSION: ['3.8']
     steps:
       - uses: actions/checkout@v3
         with:
@@ -16,7 +16,7 @@ jobs:
 
       - uses: conda-incubator/setup-miniconda@v2
         with:
-          python-version: ${{ matrix.PYTHON_VERSION }}
+          # python-version: ${{ matrix.PYTHON_VERSION }}
           miniforge-variant: Mambaforge
           miniforge-version: 4.11.0-2
           use-mamba: true
@@ -26,14 +26,14 @@ jobs:
       - name: Build
         shell: bash -l {0}
         # It seems as though flit publish expects a .pypirc file as compared to
-        # only a token passed as an argment.
+        # only a token passed as an argument.
         # See https://flit.pypa.io/en/latest/cmdline.html#flit-publish
         run: |
-          flit build --format sdist
+          flit build
 
       - uses: actions/upload-artifact@v3
         with:
-          path: dist/*.tar.gz
+          path: dist/*
 
   upload_pypi:
     name: Upload to PyPI
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -5,10 +5,6 @@ jobs:
   build_sdist:
     name: Build source distribution
     runs-on: ubuntu-latest
-    strategy:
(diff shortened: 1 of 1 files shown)


kklein:
I just realized that, sadly, as of now, github does not have tag protection. :(
https://github.com/community/community/discussions/5529

.github/workflows/build_and_publish.yaml
```diff
@@ -0,0 +1,48 @@
...
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          python-version: ${{ matrix.PYTHON_VERSION }}
```
ivergara, line 15: Why does this variable expansion work? I don't see any `matrix` attribute defined for the job.

.github/workflows/build_and_publish.yaml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          flit build --format sdist
+
+      - uses: actions/upload-artifact@v3
+        with:
+          path: dist/*.tar.gz
+
+  upload_pypi:
+    name: Upload to PyPI
+    needs: [build_sdist, upload_testpypi]
```
ivergara:
```suggestion
    needs: [build_sdist]
```

.github/workflows/build_and_publish.yaml
```diff
@@ -5,6 +5,10 @@ jobs:
   build_sdist:
     name: Build source distribution
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        PYTHON_VERSION: ['3.8', '3.9', '3.10']
```
ivergara, line 11: Unfortunately, this doesn't make sense. You just need to build once and publish, not to build 3 times and upload 3. In particular with a package that is architecture independent.

Just hardcode 3.8.

ivergara (changes requested):

.github/workflows/build_and_publish.yaml
kklein: Any clue what is breaking with the hardcoded 3.8?

.github/workflows/build_and_publish.yaml
```diff
@@ -0,0 +1,48 @@
...
+          environment-file: environment.yml
+          activate-environment: datajudge
+
+      - name: Build
+        shell: bash -l {0}
+        # It seems as though flit publish expects a .pypirc file as compared to
+        # only a token passed as an argment.
+        # See https://flit.pypa.io/en/latest/cmdline.html#flit-publish
+        run: |
+          flit build --format sdist
```
ivergara, line 28: Any particular reason why building an sdist and not a wheel?

ivergara (approved):

.github/workflows/build_and_publish.yaml
ivergara: This is very odd, I was going to say that you might need to give it as a string, but I see you did that and still was failing. Furthermore, even with subsequent changes, it's still broken and the error is not very helpful.

.github/workflows/build_and_publish.yaml
kklein: Building both now, good catch

.github/workflows/build_and_publish.yaml
```diff
@@ -2,21 +2,18 @@ name: Build and upload to PyPI
 on: [push]
 
 jobs:
-  build_sdist:
+  build_dist:
     name: Build source distribution
```
ivergara, line 6: Just `build` or perhaps `build_artifacts`

ivergara (approved):

PR #63, opened 2022-10-10
Privileges

kklein:
Illustrational branch - not actually a PR and not intended to be merged as is.

### DIFF
--- a/local.py
+++ b/local.py
@@ -68,3 +68,19 @@ with engine_jim.connect() as con:
             print("missing privilege")
         else:
             raise error
+
+
+with engine.connect() as con:
+    result = con.execute(
+        "SELECT table_name, grantee, privilege_type FROM information_schema.table_privileges WHERE grantee = 'jim'"
+    ).fetchall()
+print(result)
+
+# This outputs:
+# [('companies', 'jim', 'INSERT'),
+#  ('companies', 'jim', 'SELECT'),
+#  ('companies', 'jim', 'UPDATE'),
+#  ('companies', 'jim', 'DELETE'),
+#  ('companies', 'jim', 'TRUNCATE'),
+#  ('companies', 'jim', 'REFERENCES'),
+#  ('companies', 'jim', 'TRIGGER')]
--- a/local.py
+++ b/local.py
@@ -62,7 +62,7 @@ table_secret = sa.Table("companies_secret", metadata, autoload_with=engine_jim)
 
 with engine_jim.connect() as con:
     try:
-        result2 = con.execute("SELECT stupid FROM companies").fetchall()
+        result2 = con.execute("SELECT * FROM companies").first()
     except sa.exc.ProgrammingError as error:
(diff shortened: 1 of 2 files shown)


local.py
```diff
@@ -0,0 +1,70 @@
...
+
+metadata = sa.MetaData(engine_jim)
+
+# Both of these succeed. Even though jim doesn't have rights to read from
+# 'companies_secret', all of its metadata around columns, constraints etc.
+# can be investigaed.
+table = sa.Table("companies", metadata, autoload_with=engine_jim)
+table_secret = sa.Table("companies_secret", metadata, autoload_with=engine_jim)
+
+with engine_jim.connect() as con:
```
kklein, line 63: @svengiegerich
I looked through methods and attributes of the `table_secret` object and didn't happen to find anything evidently indicative of access rights.
Yet, as a consolation, you might use a similar approach to what is outlined here. Wrapping this code in a `can_read` function returning `False` in the "missing privilege" case and true if the `try` succeeds could potentially - albeit not beautifully - do the trick.
Likely other drivers for other databases will have error titles other than `psycopg2.errors.InsufficientPrivilege`. In how far you can find a common denominator I can't quite tell.

kklein: One could try to figure out what the according error titles for other read permissions are and check for their names in slightly altered attempts, too.

local.py
```diff
@@ -0,0 +1,86 @@
...
+    except sa.exc.ProgrammingError as error:
+        if "psycopg2.errors.InsufficientPrivilege" in str(error):
+            print("missing privilege")
+        else:
+            raise error
+
+
+with engine.connect() as con:
+    result = con.execute(
+        "SELECT table_name, grantee, privilege_type FROM information_schema.table_privileges WHERE grantee = 'jim'"
```
kklein, line 75: @svengiegerich 
A quick google search makes it look as though this schema and table were available across many dialects/dbms.
Afaik sqlalchemy relies on it to do inspection.

kklein: > USE my_db
GO
SELECT *
FROM INFORMATION_SCHEMA.TABLE_PRIVILEGES

Note that you are using `my_db` when looking for `FROM INFORMATION_SCHEMA.TABLE_PRIVILEGES`. I don't know about this but I could very well imagine this schema to live in `tempdb` on mssql. :)

kklein: At first sight it does indeed look like that. See this reference in sqlalchemy:
https://github.com/sqlalchemy/sqlalchemy/blob/d305f0130dd1cf77f8ffe9f22e63dd770630ec81/lib/sqlalchemy/dialects/mssql/information_schema.py#L107

kklein: More context on access to `INFORMATION_SCHEMA` on mssql:

> The visibility of the metadata in information schema views is limited to securables that a user either owns or on which the user has been granted some permission.

https://learn.microsoft.com/en-us/sql/relational-databases/system-information-schema-views/system-information-schema-views-transact-sql?view=sql-server-ver16#permissions

PR #64, opened 2022-10-11
Bigquery support

ClaudioVanoniQC:
This PR add basic support for BigQuery backend. 
Main changes:
- uses sqlalchemy-bigquery package
- add patches to re-map the UNION, INTERSECT and EXCEPT to the corresponding DISTINCT version which is a required keyword in BigQuery
- use dialect specific date_diff syntax
- fix under-determined "order by" function for the creation of the ranking column, which previously returned non deterministic results
- adapt tests on column capitalization
- adapt test on uniqueness by using the mod functions instead of "%"
- skip tests on regex which is not supported by current sqlalchemy-bigquery
- skip tests on primary keys not supported by BigQuery
- skip tests based on random_normal_table which takes way to long to create

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,13 +7,14 @@
 Changelog
 =========
 
-1.2.0 - 2022.xx.xx
+1.2.0 - 2022.10.21
 ------------------
 
 **New features**
 
 - Implemented specification of number of counterexamples in :meth:`~datajudge.WithinRequirement.add_varchar_regex_constraint`.
 - Implemented in-database regex matching for some dialects via ``computation_in_db`` parameter in :meth:`~datajudge.WithinRequirement.add_varchar_regex_constraint`.
+- Added support for BigQuery backends.
 
 
 **Bug fix:**


tests/integration/conftest.py
```diff
@@ -33,15 +33,21 @@ def get_engine(backend) -> sa.engine.Engine:
...
+
+    return engine
 
 
 @pytest.fixture(scope="module")
 def engine(backend):
     engine = get_engine(backend)
     with engine.connect() as conn:
-        if engine.name == "postgresql":
+        if engine.name == "postgresql" or engine.name == "bigquery":
```
ivergara, line 50: Maybe?

```suggestion
        if engine.name in ("postgresql", "bigquery"):

```

tests/integration/conftest.py
```diff
@@ -741,8 +760,9 @@ def cross_cdf_table2(engine, metadata):
 def pytest_addoption(parser):
     parser.addoption(
         "--backend",
-        choices=(("mssql", "mssql-freetds", "postgres", "snowflake")),
+        choices=(("mssql", "mssql-freetds", "postgres", "snowflake", "bigquery")),
         help="which database backend to use to run the integration tests",
+        default="bigquery",
```
ivergara, line 765: We don't play favorites :p

```suggestion

```

src/datajudge/db_access.py
```diff
@@ -24,10 +25,46 @@ def is_snowflake(engine: sa.engine.Engine) -> bool:
...
 
+def is_bigquery(engine: sa.engine.Engine) -> bool:
+    return engine.name == "bigquery"
+
+
 def get_table_columns(table, column_names):
     return [table.c[column_name] for column_name in column_names]
 
 
+def apply_patches(engine: sa.engine.Engine):
```
ivergara, line 36: Interesting.

My only concern is that this function has to be called manually for people using `bigquery`. In the tests, you're taking care of that, but users won't be directly aware.

For now, since we don't have a section with backend-specific comments, you could write in the documentation at the bottom of [this section](https://datajugde.readthedocs.io/en/latest/getting_started.html#creating-a-specification) what to do for `bigquery`.

Furthermore, our `pytest` "integration" could also handle it automatically by adding to [this function](https://github.com/Quantco/datajudge/blob/8ab90bdb291db44534d4447df85a00fc4025261a/src/datajudge/pytest_integration.py#L23) an implicit call to the `apply_patch` function before the `datajudge_engine` fixture is used.

What's your take @kklein?

ivergara (commented):
Besides a few minor issues, it looks great! Thank you for your work.

I'll let @kklein have the last word.

tests/integration/conftest.py
ClaudioVanoniQC: that was wrongly added, will remove it. I couldn't find another way in vscode to set it....Maybe you know how to do it e.g. in lauch.json?

ivergara: I guessed it was part of your workflow. 

I never tried to customize how the tests are executed with the integrated test executioner in VSCode.

ClaudioVanoniQC:
Btw, code coverage decreased but is related to the missing BigQuery backend in cicd, right?

ivergara:
> Btw, code coverage decreased but is related to the missing BigQuery backend in cicd, right?

I would say so, yes.

tests/integration/conftest.py
```diff
@@ -33,15 +33,21 @@ def get_engine(backend) -> sa.engine.Engine:
         password = os.environ.get("SNOWFLAKE_PASSWORD")
         account = os.environ.get("SNOWFLAKE_ACCOUNT", "")
         connection_string = f"snowflake://{user}:{password}@{account}/datajudge/DBO?warehouse=datajudge&role=accountadmin"
+    elif "bigquery" in backend:
+        gcp_project = os.environ.get("GOOGLE_CLOUD_PROJECT", "scratch-361908")
```
kklein, line 37: @ivergara Do you expect us to still use these identifiers once CI is set up?

src/datajudge/db_access.py
kklein: I think you raise great points @ivergara .

> but users won't be directly aware.

As you also point out, users will likely start off with some code samples from the documentation - indicating how the engine object has to be made accessible for tests. I figure that adding a comment in said documentation to call this function should be a good start. 

> Furthermore, our pytest "integration" could also handle it automatically by adding to [this function](https://github.com/Quantco/datajudge/blob/8ab90bdb291db44534d4447df85a00fc4025261a/src/datajudge/pytest_integration.py#L23) an implicit call to the apply_patch function before the datajudge_engine fixture is used.

I don't have an opinion on which of both options to pursue.

I think in general we might want to keep an eye on this topic (and potentially open issues at https://github.com/googleapis/python-bigquery-sqlalchemy) and think of ways to make the usage easiest for an end user. Yet, the suggested approach looks perfectly fine to me to get started with BigQuery. :)

tests/integration/test_column_capitalization.py
```diff
@@ -13,7 +13,9 @@ def test_column_existence(
 ):
     if is_mssql(engine) and use_uppercase_column != use_uppercase_query:
         pytest.skip("Mssql interface expects exact capitalization.")
-    if is_postgresql(engine) and use_uppercase_query:
+    if is_bigquery(engine) and use_uppercase_column != use_uppercase_query:
+        pytest.skip("BigQuery interface expects exact capitalization.")
+    if is_postgresql(engine) and True:
```
kklein, line 18:
```suggestion
    if is_postgresql(engine):
```

kklein (commented):
Thanks so much Claudio! :)

tests/integration/conftest.py
ivergara: No, but it doesn't hurt to keep them there as placeholders.

ClaudioVanoniQC: ok will leave it then, also because the name sounds generic and anonym

tests/integration/test_column_capitalization.py
ClaudioVanoniQC: removed the unnecessary "and True"

src/datajudge/db_access.py
ClaudioVanoniQC: Very good point, I missed that. I like @ivergara 's suggestion to add the patch before using the datajudge_engine fixture. Alternatively, one could have a second "engine" fixture which return a patched datajudge_engine one. But since the patch should only be necessary until sqlalchemy fix the issues, I wouldn't go for it.

kklein (approved):
Lgtm - thanks! :)
I'd appreciate it if you could run the BigQuery integration tests locally with the latest state of the code since we don't have them in CI yet.

kklein:
Many thanks!

PR #66, opened 2022-10-21
add: infer primary columns from table definition

YYYasin19:
I am somewhat confused, thought the implementation was even easier but totally forgot that we have a clear seperation between test design time and test execution time (which I find really cool, btw).

Does anyone have an idea for a cleaner solution?

### DIFF
--- a/src/datajudge/constraints/miscs.py
+++ b/src/datajudge/constraints/miscs.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import List, Optional, Set, Tuple
 
 import sqlalchemy as sa
@@ -76,9 +77,14 @@ class Uniqueness(Constraint):
         # only check for primary keys when actually defined
         # otherwise default back to searching the whole table
         if self.infer_pk_columns and (
-            columns := db_access.get_primary_keys(engine, self.ref)[0]
+            pk_columns := db_access.get_primary_keys(engine, self.ref)[0]
         ):
-            self.ref.columns = columns
+            self.ref.columns = pk_columns
+            if not pk_columns:  # there were no primary keys found
+                warnings.warn(
+                    f"""No primary keys found in {self.ref.get_string()}.
+                    Uniqueness will be tested for all columns."""
+                )
 
         unique_count, unique_selections = db_access.get_unique_count(engine, self.ref)
         row_count, row_selections = db_access.get_row_count(engine, self.ref)
--- a/src/datajudge/constraints/miscs.py
+++ b/src/datajudge/constraints/miscs.py
@@ -53,7 +53,7 @@ class Uniqueness(Constraint):
         ref: DataReference,
         max_duplicate_fraction: float = 0,
         max_absolute_n_duplicates: int = 0,
-        infer_pk_columns=False
+        infer_pk_columns=False,
     ):
         if max_duplicate_fraction != 0 and max_absolute_n_duplicates != 0:
             raise ValueError(
@@ -72,10 +72,12 @@ class Uniqueness(Constraint):
         super().__init__(ref, ref_value=ref_value)
 
     def test(self, engine: sa.engine.Engine) -> TestResult:
-        
+
         # only check for primary keys when actually defined
         # otherwise default back to searching the whole table
-        if self.infer_pk_columns and (columns := db_access.get_primary_keys(engine, self.ref)[0]):
+        if self.infer_pk_columns and (
+            columns := db_access.get_primary_keys(engine, self.ref)[0]
+        ):
             self.ref.columns = columns
 
         unique_count, unique_selections = db_access.get_unique_count(engine, self.ref)
--- a/src/datajudge/constraints/miscs.py
+++ b/src/datajudge/constraints/miscs.py
@@ -53,6 +53,7 @@ class Uniqueness(Constraint):
         ref: DataReference,
         max_duplicate_fraction: float = 0,
         max_absolute_n_duplicates: int = 0,
+        infer_pk_columns=False
     ):
         if max_duplicate_fraction != 0 and max_absolute_n_duplicates != 0:
             raise ValueError(
@@ -66,9 +67,17 @@ class Uniqueness(Constraint):
             ref_value = ("absolute", max_absolute_n_duplicates)
         else:
             ref_value = ("relative", 0)
+
+        self.infer_pk_columns = infer_pk_columns
         super().__init__(ref, ref_value=ref_value)
 
     def test(self, engine: sa.engine.Engine) -> TestResult:
+        
+        # only check for primary keys when actually defined
+        # otherwise default back to searching the whole table
+        if self.infer_pk_columns and (columns := db_access.get_primary_keys(engine, self.ref)[0]):
+            self.ref.columns = columns
+
         unique_count, unique_selections = db_access.get_unique_count(engine, self.ref)
         row_count, row_selections = db_access.get_row_count(engine, self.ref)
         self.factual_selections = row_selections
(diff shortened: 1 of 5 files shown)


kklein:
Looks great to me, thanks for working on this! :)

In addition to the actual implementation, we generally try to have a test for each parameter.

Here[0] we currently have a test for `add_uniqueness_constraint`, which relies on `mix_table2`[1]. One could add a test, e.g. `test_uniqueness_within_inferred`, on a new fixture with the same data but an actual primary key definition. This is fairly convenient with `sqlalchemy`, e.g.
```python
    columns = [
        sa.Column("col_int", sa.Integer(), primary_key=True),
        sa.Column("col_varchar", sa.String()),
        sa.Column("col_date", sa.DateTime()),
    ]
```
instead of
```python
    columns = [
        sa.Column("col_int", sa.Integer()),
        sa.Column("col_varchar", sa.String()),
        sa.Column("col_date", sa.DateTime()),
    ]
```

Let me know what you think @YYYasin19 !

[0] https://github.com/Quantco/datajudge/blob/main/tests/integration/test_integration.py#L1612-L1649
[1] https://github.com/Quantco/datajudge/blob/main/tests/integration/conftest.py#L118-L136

src/datajudge/constraints/miscs.py
```diff
@@ -66,9 +67,19 @@ def __init__(
...
+
+        self.infer_pk_columns = infer_pk_columns
         super().__init__(ref, ref_value=ref_value)
 
     def test(self, engine: sa.engine.Engine) -> TestResult:
+
+        # only check for primary keys when actually defined
+        # otherwise default back to searching the whole table
+        if self.infer_pk_columns and (
+            columns := db_access.get_primary_keys(engine, self.ref)[0]
```
ivergara, line 79: This will fail since we're still testing against Python 3.7.

YYYasin19:
Nice idea, I think I'm going to go that way.

I'd like to split testing the functionality in two parts:
1. is the function parameter `infer_pk_columns` used correctly?
2. does `db_access.get_primary_keys` work correctly?

Is the second one already tested somewhere? Couldn't find anything on the first glance.

kklein:
Sgtm!

> Is the second one already tested somewhere? Couldn't find anything on the first glance.

I don't think there is a test for it in isolation. Yet, it is tested as part of some integration tests, e.g. here:
https://github.com/Quantco/datajudge/blob/main/tests/integration/test_integration.py#L1593-L1609

YYYasin19:
I am not that satisfied with my current test, but unfortunately the following does not work (and is not very beautiful, either):
```py
@pytest.mark.parametrize("data", [
    (identity, "col_int"),
    (negation, "col_date"),
    (identity, "col_int,col_date")
])
def test_uniqueness_within_infer_pk(engine, mix_table2, data):
    # We purposefully select a non-unique column ["col_date"] to validate
    # that the reference columns are overwritten.
    operation, columns_to_pk = data

    # alter columns such that they contain the primary key
    _, schema, table = mix_table2
    try:
        engine.execute(f"ALTER TABLE {schema}.{table} DROP PRIMARY KEY")
    except Exception as e:
        print(e)
    engine.execute(f"ALTER TABLE {schema}.{table} ADD PRIMARY KEY ({columns_to_pk})")

    req = requirements.WithinRequirement.from_table(*mix_table2)
    req.add_uniqueness_constraint(columns=columns_to_pk.split(","), infer_pk_columns=True)
    test_result = req[0].test(engine)
    assert operation(test_result.outcome), test_result.failure_message
```
because `db_access.get_primary_keys()` is cached at application start, which is fine, since usually the schema should not change during the test.

Therefore the current implementation uses a small pytest trick to prevent the test from being even more verbose.
I also added a small note to the documentation, that when `columns` and `infer_pk` is not given, we fall back to comparing rows as whole.
Unfortunately, this behavior is somewhat dangerous for our tests: If the extraction if primary keys fails, we check that each row is unique, which is true in all tables that have an implicit primary key / unique column.

kklein:
> I also added a small note to the documentation, that when columns and infer_pk is not given, we fall back to comparing rows as whole.

Great, thank you!

> Unfortunately, this behavior is somewhat dangerous for our tests: If the extraction if primary keys fails, we check that each row is unique, which is true in all tables that have an implicit primary key / unique column.

I see. Could we add a table with very few entries to account for this?

kklein:
On a different note: Do you have a clue on why the commits of this PR don't seem to have the usual GitHub actions run against them? Is this due to this PR coming from a fork, rather than a datajudge branch?

YYYasin19:
> Is this due to this PR coming from a fork, rather than a datajudge branch?

I think so, yes. You might need to head to the [repo settings here](https://github.com/Quantco/datajudge/settings/actions) and allow actions for fork runs as well. But I'm not exactly sure.

YYYasin19:
> I see. Could we add a table with very few entries to account for this?

Could you elaborate more on what you mean here?

The only thing I was able to come up with right now is to just validate the result within the data reference, since the values are updated during test time.
(I have to admit, I don't like the code style that much since it modifies some state about the reference in the background.. stateful code always seems to be hard to refactor. good thing we have tests :) )

kklein:
> I think so, yes. You might need to head to the [repo settings here](https://github.com/Quantco/datajudge/settings/actions) and allow actions for fork runs as well. But I'm not exactly sure.

Thanks for the reference! I adapted it and would hope that the actions will be triggered with your next commit. Let's see!

> Could you elaborate more on what you mean here?

I'm sorry - upon thinking about it once more I come to the conclusion that my statement doesn't make sense. To get back to your point

>  [...] If the extraction of primary keys fails [...]

I suppose that the 'failure mode' you are referring to here is that `None` or an empty list is returned by `get_primary_keys`.
Could we add a warning for when this is the case?
In our test, we could then assert that no warning has been raised in the process.

YYYasin19:
> I suppose that the 'failure mode' you are referring to here is that None or an empty list is returned by get_primary_keys.
Could we add a warning for when this is the case?
In our test, we could then assert that no warning has been raised in the process.

Nice idea! The user should be aware that the table might be ill-defined. I've added that.
I am, however, not in favor of testing warnings, since it seems to be somewhat messy, depends on some level, etc.
What do you think of the current solution, i.e. additionally testing that the correct columns were extracted?

kklein:
> I've added that. 

Cool, thanks! :)

> I am, however, not in favor of testing warnings, since it seems to be somewhat messy, depends on some level, etc.

I see.

> What do you think of the current solution, i.e. additionally testing that the correct columns were extracted?

Lgtm!

kklein:
However, I still don't see any further GitHub actions being triggered. :(

I'll ask around to see if someone has an idea on how to fix this. If you want to move fast(er), I would suggest to open up a datajudge branch since I sadly don't have any short term fix to offer.

YYYasin19:
CI is now running after rebase 🥳 
If you're fine with my release notes, this should be finished, as far as I can see :-)

YYYasin19:
So, checking out the integration test results [from the working CI PR](https://github.com/Quantco/datajudge/actions/runs/3372032846/jobs/5594985679), I noticed that the test fails for regular relational databases, i.e. mssql and postgres, since these do not allow PK values. The issue lies [here](https://github.com/YYYasin19/datajudge/blob/65f5cc3b8598b152ba7890d51026a1dde9b02d67/tests/integration/conftest.py#L165).

What would you recommend doing here? Remove the negation test? Somehow disable primary key checking afterwards (huge hack, imo).

kklein:
I guess ideally we would mock the relevant objects. I would be ok with getting rid of this particular test. If it ends up causing problems in use, we can revisit it, too.

YYYasin19:
thanks! @kklein and @ivergara

PR #71, opened 2022-11-02
Restrict push worfklow trigger

kklein:


xhochy (approved):
This would break bots like the pre-commit-autoupdate bot. Here CI would only start on the `push` event and not on the `pull_request` event due to Github's loop prevention mechanism. We can workaround that by whitelisting the respective branches in the `ci.yaml` once we add this workflow here, too (we should do it!)

PR #79, opened 2022-11-04
Bigquery ci

ivergara:


tests/integration/conftest.py
```diff
@@ -34,8 +34,8 @@ def get_engine(backend) -> sa.engine.Engine:
         account = os.environ.get("SNOWFLAKE_ACCOUNT", "")
         connection_string = f"snowflake://{user}:{password}@{account}/datajudge/DBO?warehouse=datajudge&role=accountadmin"
     elif "bigquery" in backend:
-        gcp_project = os.environ.get("GOOGLE_CLOUD_PROJECT", "scratch-361908")
-        connection_string = f"bigquery://{gcp_project}"
+        # gcp_project = os.environ.get("GOOGLE_CLOUD_PROJECT", "scratch-361908")
```
kklein, line 37: Do we want to keep this?

kklein (approved):

tests/integration/conftest.py
ivergara: It's not needed. 

Before merging this, I want to see how much the time would improve if we use `pytest-xdist` to parallelize the test execution in BigQuery (maybe helps with Snowflake too).

kklein:
> Before merging this, I want to see how much the time would improve if we use pytest-xdist to parallelize the test execution in BigQuery (maybe helps with Snowflake too).

I played around with it a bit. After switching the fixture scopes from `"module"` to `"session"`, the runtime of the snowflake tests went down quite a bit by using `-n auto`. Looks promising!

ivergara:
A non-scientific "test"
                                xdist
                   without        with
snowflake: 5m 49s        3m 2s
bigquery:    20+m         6m 22s

PR #85, opened 2022-11-09
Provide custom names for tests

kklein:


kklein:
fyi @jonashaag @YYYasin19

jonashaag:
🎉

tests/unit/test_pytest.py
```diff
@@ -0,0 +1,60 @@
...
+def test_collect_data_tests():
+    req = WithinRequirement.from_raw_query("select * from example", "example_table")
+    req.add_n_rows_equality_constraint(42)
+    test_func = pytest_integration.collect_data_tests([req])
+    (parametrize_mark,) = test_func.pytestmark
+    assert parametrize_mark.name == "parametrize"
+    assert parametrize_mark.args[1] == list(req)
+    assert [parametrize_mark.kwargs["ids"](c) for c in parametrize_mark.args[1]] == [
+        "NRowsEquality::example_table"
+    ]
```
YYYasin19, line 15: Is this the same?
```py3
assert [parametrize_mark.kwargs["ids"](req) for req in list(req)] == ["NRowsEquality::example_table"]
```

YYYasin19:
Very nice! 🥳 
I think it solves the use case nicely.

tests/unit/test_pytest.py
kklein: I find `req in list(req)` a little confusing. Did you mean `[parametrize_mark.kwargs["ids"](constraint) for constraint in req]`?

If so, it should be the same if everything goes well. Using the indirection via `parametrize_mark.args` might be considered a further test whether that is actually the case.

But maybe I'm missing something.

ivergara (approved):

PR #86, opened 2022-11-11
Enable tests relying on `snowflake-sqlalchemy` 1.4.0

kklein:


ivergara (approved):
Wouldn't it be wise to give a lower bound to snowflake_sqlalchemy?

PR #96, opened 2022-12-20
Impala

kklein:


kklein:
I cannot reproduce the current issue locally. :/
https://github.com/Quantco/datajudge/actions/runs/3740746262/jobs/6349480222

kklein:
It seems to me as though after 'a while' of running the tests, a service of one of the docker containers starts failing leading to 'connection reset by peer errors, which can be seen e.g. here[0].

In order to run the tests successfully in CI, I biw simply split up the test suite into multiple checks, which seems to work.

I'm not a fan of this approach and would like to get rid of this eventually. Yet, I'm not sure I'm able to afford the time to fix this right now.

Would this approach seem acceptable to you @ivergara ?

[0] https://github.com/Quantco/datajudge/actions/runs/3930026693/jobs/6719529754

ivergara:
> It seems to me as though after 'a while' of running the tests, a service of one of the docker containers starts failing leading to 'connection reset by peer errors, which can be seen e.g. here[0].
> 
> In order to run the tests successfully in CI, I biw simply split up the test suite into multiple checks, which seems to work.
> 
> I'm not a fan of this approach and would like to get rid of this eventually. Yet, I'm not sure I'm able to afford the time to fix this right now.
> 
> Would this approach seem acceptable to you @ivergara ?

I'm fine with a divide-and-conquer approach.

ivergara (approved):

PR #100, opened 2023-01-13
Rework failure message for `RowMatchingEquality`

kklein:


kklein:
@jonashaag the failure message looks as follows for me on the integration tests:

```
AssertionError: 0.42857142857142857143 > 0.41 of the rows differ on a match of 7 rows between tempdb.dbo.row_match_table and tempdb.dbo.row_match_table2. Matched on ['col_match1', 'col_match2'] and ['col_match1', 'col_match2']. Compared on ['col_compare1', 'col_compare2'] and ['col_compare1', 'col_compare2'].
```

Yet, the indication of the matched columns is not something I introduced in this PR. As a matter of fact, it should've already been present beforehand. Would you mind looking into why this doesn't seem to be the case for you?

jonashaag:
Awesome! It could very well be that it was simply cut by pytest and I would've had to use `-v` or something

jonashaag (approved):

PR #101, opened 2023-01-17
Test drive new Micromamba pre-commit integration

jonashaag:
Taking this repo as a guinea pig


jonashaag:
Ok this works but let's do it in the cookiecutter

PR #107, opened 2023-02-24
Add experimental support for db2.

kklein:


tests/integration/conftest.py
```diff
@@ -111,7 +113,7 @@ def mix_table1(engine, metadata):
     table_name = "mix_table1"
     columns = [
         sa.Column("col_int", sa.Integer()),
-        sa.Column("col_varchar", sa.String()),
+        sa.Column("col_varchar", sa.String(40)),
```
ivergara, line 116: I wouldn't have expected this to break `impala`'s integration tests.

kklein: Haha I wasn't exactly hoping for this either. I'll try to figure something out... Worst case we need a case distinction.

kklein:
FYI @FlorianOttQC I hope to have experimental support for db2 released soon.

kklein:
> I'm kind of surprised that the tests didn't break with sqlalchemy 2.0.

I would suppose that we have some dependencies expecting sqlalchemy < 2.0

PR #112, opened 2023-03-07
Implement a `NumericPercentile` `Constraint`

kklein:


src/datajudge/constraints/numeric.py
```diff
@@ -158,3 +158,55 @@ def test(self, engine: sa.engine.Engine) -> TestResult:
...
+        ref2: DataReference = None,
+        expected_percentile: float = None,
+    ):
+        super().__init__(
+            ref,
+            ref2=ref2,
+            ref_value=expected_percentile,
+            name=name,
+        )
+        if k < 0 or k > 100:
```
ivergara, line 181:
```suggestion
        if 0 > k > 100:
```

kklein: I guess I accepted to quickly :D

ivergara: Or I suggested too quickly. I'm missing a `not` 

```if not (0 > k > 100):```

**Comment by kklein** [comment]
`failure_message` now looks as such:
```The 20-th percentile of tempdb.dbo.int_table1's column(s) 'col_int' was expected to be 2.8 but was 3, resulting in an absolute difference of 0.20000000000000018. The maximally allowed absolute deviation would've been 0.```

**Comment by kklein** [comment]
I'm not sure why codecov doesn't seem to pick up the coverage gain achieved via the unit tests.

File: src/datajudge/db_access.py [hunk 2]
```diff
@@ -777,6 +777,39 @@ def column_operator(column):
     return get_column(engine, ref, aggregate_operator=column_operator)
 
 
+def get_kth_percentile(engine, ref, k):
+    row_count = "row_count"
+    row_num = "row_num"
```
ivergara, line 782:
```suggestion
    ROW_COUNT = "dj_row_count"
    ROW_NUM = "dj_row_num"
```
Two changes
1. Capitalized, to show the intention that these strings are a constant.
2. Use the prefix `dj` to make it explicit that these are arbitrary ones from datajudge.

I find the first more relevant than the second. I'm open to keeping them as they're given a good argument.

src/datajudge/constraints/numeric.py
```diff
@@ -158,3 +158,79 @@ def test(self, engine: sa.engine.Engine) -> TestResult:
...
+    def __init__(
+        self,
+        ref: DataReference,
+        k: float,
+        max_absolute_deviation: Optional[float] = None,
+        max_relative_deviation: Optional[float] = None,
+        name: Optional[str] = None,
+        *,
+        ref2: DataReference = None,
+        expected_percentile: float = None,
```
ivergara, line 173: I wonder why `mypy` doesn't complain from this incongruency between annotation and default value. Now, me personally, seems odd default value. Looking into the code in `db_access` I see that an `expected_percentile` is always given. 

Any reason not to move this parameter before the `*` and not ake it default to `None`?

And since I'm on this, the previous line about `ref2` also seems like a typing issue which should be `ref2: DataReference | None = None`. If this is happening in other `Constraint`s then I apologize for that oversight in the past.

src/datajudge/constraints/numeric.py
```diff
@@ -158,3 +158,79 @@ def test(self, engine: sa.engine.Engine) -> TestResult:
         )
         result = deviation <= self.max_absolute_deviation
         return TestResult(result, assertion_text)
+
+
+class NumericPercentile(Constraint):
+    def __init__(
+        self,
+        ref: DataReference,
+        k: float,
```
ivergara, line 167: Minor nitpicking, I never liked the name just being `k` in here. Now, when reading further down some docstring where it mentions the `k-th` percentile, it made sense, but it lacks context here. The main reason is that we decided not to document in docstrings here but in the convenience functions.

ivergara (approved):

src/datajudge/constraints/numeric.py
kklein: I first couldn't come up with a better technical term for it. But I actually think we could just call it `percentage`. :)

ivergara: Or perhaps just `percentile`?

src/datajudge/constraints/numeric.py
kklein: I'm a little confused about this, too. I believe to vaguely recall that we once thought this would mean that mypy accepts the keyword argument not being used by the caller, but would complain about explicitly setting it to `None`. I might be wrong, though.

> Looking into the code in db_access I see that an expected_percentile is always given.

The function `get_percentile` in `db_access` retrieves the percentile of a given data source. If we use this constraint in a `WithinRequirement`, we expect an an `expected_percentile` to be given - since we need something to compare the factual percentile of this single data source to. If we use this constraint in a `BetweenRequirement`, we don't need a fixed reference value `expected_percentile` but instead can just compare the percentiles of both data sources. Therefore this is an optional parameter.

src/datajudge/db_access.py
kklein: I'm fully with you on the second point!

Regarding the first point, PEP8 [0] says:

> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.

I would've read this as if it would be more appropriate to lower-case the aforementioned objects. Do you disagree?

[0] https://peps.python.org/pep-0008/#constants

src/datajudge/constraints/numeric.py
kklein: I think that wouldn't be what we want. The way I understand the terminology it holds that:
$\Pr[x <= percentile] = percentage / 100$

where there is incongruency wrt the strictness of the inequality.

src/datajudge/constraints/numeric.py
kklein: Found the culprit!

https://github.com/Quantco/datajudge/blob/main/pyproject.toml#L58

If you don't like this too much, should we move it into a discussion in a separate GitHub issue?

PR #113, opened 2023-03-10
Implement a ``Constraint`` limiting the fraction of ``NULL`` values in a column.

kklein:
In addition to introducing a ``Constraint`` limiting the fraction of NULL values in a column, this PR removes the pre-existing ``NullAbsence`` constraint since the latter is a special case of the former.

See https://github.com/Quantco/datajudge/pull/113/commits/c0348ee4a91007487f08a2078df57cc8dc46df69 for the removal of ``NullAbsence``.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,16 @@
 Changelog
 =========
 
+
+1.5.0 - 2022.03.XX
+------------------
+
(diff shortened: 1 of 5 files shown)


ivergara (approved):
Same concerns as in #112 `__init__` in the constraint.

PR #125, opened 2023-04-12
add: parallel testing for sqlalchemy 2.0 + refactored CI

YYYasin19:
First try at testing sqlalchemy>2.0 and sqlalchemy<2.0 in one CI.

The idea is that our environment.yml should not include any vendor-specific package. The CI will install those packages required for certain parts by passing them to the micromamba GitHubAction.

I also cleaned up some of the names to make them more readable in the web view; that's purely optional, though. :)


YYYasin19:
Also, I switched from miniconda (with mamba) to micromamba which is even faster.
As a comparison: In GitHub Actions `setup-micromamba` takes 30s for our environment and the prev. `miniconda` action took 1min 30s.

YYYasin19:
What I don't like about this solution is that we now have ~50 checks that need to run every time.
Can we:
1. remove some Python versions, e.g. either the newest (since we're testing the older ones and should have backwards-compatibility) or the oldest (to drop support)
2. run tests based on PR labels, maybe? `if: ${{ github.event.label.name == 'label_name' }}` would allow us to run only unit tests on "pure" PRs and integration tests depending on the label. Before merge, we could then run _all_ tests (by adding another label "ready for merge")
3. ???

environment.yml
```diff
@@ -19,11 +19,7 @@ dependencies:
   - sqlalchemy
   - psycopg2
   - freetds
-  - snowflake-sqlalchemy>=1.4.0
-  - sqlalchemy-bigquery
   - flit-core
   - flit
   - sphinx-autodoc-typehints
   - impyla
```
ivergara, line 25: This should have to go too.

YYYasin19:
> Thank you! Exactly what I was thinking we should be doing but didn't have the clear path forward!

♥️

> There are a few that will always fail for a while because the environment won't be satisfiable. We can already remove the check for sqlalchemy 2.0+ for those. Or make them warnings but not failures.

Okay yes we should probably remove those from the matrix directly and only test for sqlalchemy.

>  I'd be fine with being more aggressive here.

Maybe just 3.10 and 3.11 then from now on forwards?

> I do like this idea. And would be a label per backend we support?

Yes, exactly!
And then, when the PR is marked as "Ready for Review" (either with that button or with another label "`review`") we let all of them run for every commit.

YYYasin19:
Okay, the following comments detail all my changes so far and I'd like your opinion on that!

.github/actions/pytest/action.yml
```diff
@@ -0,0 +1,27 @@
+name: pytest
+description: installs package, runs pytest and creates coverage report.
+
+inputs:
+  backend:
+    description: which database backend to use
+    required: false
+  args:
+    description: arguments that are passed to pytest, e.g. "-k filter" or "tests/integration"
+    required: false
```
YYYasin19, line 10: I've added a new action that runs our tests based on an argument
The goal was to save repetitive code here, but the end result is just a few lines shorter on the main `ci.yml` file

.github/workflows/ci.yaml
```diff
@@ -28,47 +28,44 @@ jobs:
...
     runs-on: ubuntu-latest
     env:
       CI: True
     strategy:
       fail-fast: true
       matrix:
-        PYTHON_VERSION: ['3.8', '3.9', '3.10']
+        # PYTHON_VERSION: ['3.8', '3.9', '3.10']
+        PYTHON_VERSION: ['3.10']
+        SA_VERSION: ['<=2.0', '>=2.0']
```
YYYasin19, line 41: there's now an `SA_VERSION` matrix element for most backends (where it makes sense) that also tests for the newer `sqlalchemy` version.
todo: we should pin this down, probably, since we can't guarantee every new version above 2.0

.github/workflows/ci.yaml
```diff
@@ -28,47 +28,44 @@ jobs:
...
+        uses: ./.github/actions/pytest
         with:
-          file: ./coverage.xml
+          args: "tests/unit"
 
-  linux-integration_tests-sqlserver:
-    name: "Linux - integration tests - Python ${{ matrix.PYTHON_VERSION }} - mssql"
+  sqlserver:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'sqlserver') || contains(github.event.pull_request.labels.*.name, 'ready') }}
+    name: "Python ${{ matrix.PYTHON_VERSION }} - SA ${{ matrix.SA_VERSION }} - mssql"
```
YYYasin19, line 60: This filter checks if the PR has a label `sqlserver` (for example) or if the PR has a label `ready` and runs the following test accordingly.
Actually saves a lot of time, more than I thought.

.github/workflows/ci.yaml
```diff
@@ -137,34 +131,31 @@ jobs:
...
-        with:
-          file: ./coverage.xml
-
+      - name: Setup conda
+        uses: mamba-org/provision-with-micromamba@v15
+        with:
+          extra-specs: |
+            python=${{ matrix.PYTHON_VERSION }}
+            sqlalchemy${{ matrix.SA_VERSION }}
+            psycopg2
```
YYYasin19, line 140: Installs the micromamba environment conditionally, i.e. for every backend and sqlalchemy version.
Therefore, these packages are missing in our default `environment.yml`.
We should probably create a new `environment-dev.yml` that has _all_ the dependencies or write a short comment that the dev has to install the dependencies for his/her backend himself later.

.github/workflows/ci.yaml
```diff
@@ -267,26 +253,24 @@ jobs:
...
+    name: "Python ${{ matrix.PYTHON_VERSION }} - SA ${{ matrix.SA_VERSION }} - Impala - pt1"
     runs-on: ubuntu-20.04
     env:
       CI: True
     strategy:
       fail-fast: false
       matrix:
         PYTHON_VERSION: [ '3.8']
+        SA_VERSION: ["<=2.0", ">=2.0"]
+        PYTEST_ARG: ["tests/integration/test_column_capitalization.py", "tests/integration/test_data_source.py", "tests/integration/test_integration.py -k row", "tests/integration/test_integration.py -k uniques", "tests/integration/test_integration.py -k date", "tests/integration/test_integration.py -k varchar", "tests/integration/test_integration.py -k numeric"]
```
YYYasin19, line 273: this might be controversial but I replaced the 7 impala jobs (they all looked the same :D) with a matrix argument that is passed to the pytest action which
just looked shorter, behaviour should be the same.

kklein: Great idea!

.github/workflows/ci.yaml
```diff
@@ -185,68 +176,63 @@ jobs:
...
     runs-on: ubuntu-latest
+    env:
+      SNOWFLAKE_USER: DATAJUDGE
+      SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+      SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
     strategy:
       fail-fast: false
       matrix:
-        PYTHON_VERSION: [ '3.8' ]
+        PYTHON_VERSION: [ '3.9' ]
```
ivergara, line 205:
```suggestion
        PYTHON_VERSION: [ '3.8' ]
```
Let's keep 3.8 here still.

.github/workflows/ci.yaml
```diff
@@ -267,26 +253,24 @@ jobs:
...
-          pytest -n auto --cov=datajudge --cov-report=xml --cov-append --backend=bigquery tests/integration
-      - name: Generate code coverage report
-        uses: codecov/codecov-action@v3.1.1
+      - name: Run tests
+        uses: ./.github/actions/pytest
         with:
-          file: ./coverage.xml
-
+          backend: bigquery
+          args: tests/integration
```
ivergara, line 260: This should have the `-n auto` argument to `pytest`, if not the test just takes ages to run.

ivergara (commented):
Once again, great job!

.github/workflows/ci.yaml
```diff
@@ -28,28 +28,28 @@ jobs:
...
     runs-on: ubuntu-latest
     env:
       CI: True
     strategy:
       fail-fast: true
       matrix:
-        PYTHON_VERSION: ['3.8', '3.9', '3.10']
+        # PYTHON_VERSION: ['3.8', '3.9', '3.10']
+        PYTHON_VERSION: ['3.10']
+        SA_VERSION: ['<=2.0', '>=2.0']
```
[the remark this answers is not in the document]
YYYasin19: Yep, good point. It would need to be `<2.0` and `>=2.0`, though, right?

YYYasin19:
Nice, the label automation works!
Everytime a new label is added, the tasks run again. Then, when the final label `ready` will be added, all tests run through.
We could also add a label action that adds the `ready` label when a user presses `Ready for review`, but that might be a little bit too much.

ivergara (approved):
Looks great to me!

kklein (approved):
Thank you!

YYYasin19:
Okay the last change is just the PR template s.t. new users know about the labeling situation.

YYYasin19:
Does anyone know why the CI fails for DB2?

ivergara:
Unfortunately not. It seems that some tests are quite flaky.

kklein:
Hmm I reran the db2 tests 3 times now - also on a different branch which is based on this one -and they failed every time. It's not obvious to me that they're flaky. Could it be that we have a dependency problem, i.e. that with the new setup, we run into an incompatibility we didn't have before?

ivergara:
> Hmm I reran the db2 tests 3 times now - also on a different branch which is based on this one -and they failed every time. It's not obvious to me that they're flaky. 

I meant flaky because we saw issues with snowflake while updating actions with dependabot.

> Could it be that we have a dependency problem, i.e. that with the new setup, we run into an incompatibility we didn't have before?

That might be a very weird corner-case, but I've seen some. It could be that for DB2, it's installing something slightly different for some reason.

YYYasin19:
do you mean incompatibility regarding the sqlalchemy version or that something in the dependencies of our (non-pinned) dependencies changed?

Is there a way to test this locally (like with Postgres or SQLServer)?

kklein:
> Is there a way to test this locally

./start_db2

pytest tests/integration --backend=db2

They are indeed failing for me locally, yet with a different error, a Segmentation fault.

YYYasin19:
It seems like I can't deploy the container on arm mac :/
Is the version still the same?

kklein:
> Is the version still the same?

According to the changelog, nothing should've changed since October:
https://hub.docker.com/r/ibmcom/db2

.github/actions/pytest/action.yml
```diff
@@ -16,6 +16,7 @@ runs:
       shell: bash -l {0}
       run: |
         source ~/.bash_profile
```
ivergara, line 18: Maybe you add a comment line explaining the need for the following two lines? Mainly for posteriority, since this will be the only ODD action in an otherwise very regular and repetitive definition structure.

kklein (approved):

.github/actions/pytest/action.yml
YYYasin19: was removed again :)

environment.yml
```diff
@@ -15,15 +15,8 @@ dependencies:
   - sphinx_rtd_theme
   - sphinxcontrib-apidoc
   - pre-commit
-  - pyodbc
   - sqlalchemy
-  - psycopg2
   - freetds
```
ivergara, line 19: Technically, this could also be removed.

ivergara (approved):
Great job once again!

PR #136, opened 2023-05-11
Add variant distribution constraint

0xbe7a:
This PR adds a VariantDistributionConstraint which checks if the distribution of values in a column falls within the specified minimum and maximum bounds.

### DIFF
--- a/src/datajudge/constraints/uniques.py
+++ b/src/datajudge/constraints/uniques.py
@@ -296,14 +296,16 @@ class NUniquesMaxGain(NUniques):
         return super().test(engine)
 
 
-class VariantDistributionConstraint(Constraint):
+class CategoricalBoundConstraint(Constraint):
     """
-    VariantDistributionConstraint is a class to check if the distribution of values in a column
-    falls within the specified minimum and maximum bounds.
+    `CategoricalBoundConstraint` is a class to check if the share of a specific value
+    in a column falls within the specified minimum and maximum bounds.
     The constraint compares the factual distribution of values in a column of a `DataSource`
-    with the target distribution supplied as a dictionary where keys represent the unique values
-    and the corresponding tuple values represent the minimum and maximum allowed shares of
-    the respective unique value in the column.
+    with the target distribution supplied as a dictionary where keys represent
+    some unique values and the corresponding tuple values represent the minimum and maximum allowed shares of
+    the respective unique value in the column. The additional `default_bounds` parameter specifies
+    the minimum and maximum bounds for all distinct values that are not explicitly outlined
+    in the target distribution dictionary.
 
     The constraint is initialized with a `DataReference` object, the target distribution,
     a minimum and maximum bound for each value not in the given distribution and
@@ -347,8 +349,6 @@ class VariantDistributionConstraint(Constraint):
 
         if violations:
             assertion_text = f"{self.ref.get_string()} contains element(s):\n"
-
-            violations = (factual - max_counts) | (min_counts - factual)
 
             for variant in violations:
                 actual_share = factual[variant] / total
--- a/src/datajudge/constraints/uniques.py
+++ b/src/datajudge/constraints/uniques.py
@@ -305,7 +305,8 @@ class VariantDistributionConstraint(Constraint):
     and the corresponding tuple values represent the minimum and maximum allowed shares of
     the respective unique value in the column.
 
-    The constraint is initialized with a `DataReference` object, the target distribution, and
+    The constraint is initialized with a `DataReference` object, the target distribution,
+    a minimum and maximum bound for each value not in the given distribution and
     optionally a name and any additional keyword arguments.
 
     Example use cases include testing for consistency in columns with expected categorical values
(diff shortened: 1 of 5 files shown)


src/datajudge/constraints/uniques.py
```diff
@@ -292,3 +294,76 @@ def compare(
     def test(self, engine: sa.engine.Engine) -> TestResult:
         self.max_relative_gain = self.max_relative_gain_getter(engine)
         return super().test(engine)
+
+
+class VariantDistributionConstraint(Constraint):
```
YYYasin19, line 299: I am not a 100% confident in the name but don't have a lot of other ideas as well.
Something along the line of `categorical` or `factor` would help one to understand it more directly, probably.

YYYasin19:
I like this a lot! 🚀 

To give a concrete example that might highlight how this could be used:
In our project, we have a timestamp column representing the birthdate of a person. When reading in data, this value gets messed up sometimes. This results in a lot of outliers (e.g. people are suddenly born in `1762`) but also a lot of misinterpreted data which causes the values to all be around ~1970. 

With this test, we could quantize all timestamps to their respective year resulting in a categorical column with values in `[1900, 1901, ..., 2023]` and then check that each category only takes up at most 5% of the column -- preventing accumulations.

src/datajudge/requirements.py
```diff
@@ -345,6 +345,63 @@ def add_n_uniques_equality_constraint(
...
+
+        `columns` is a list of column names from the `DataSource` to apply the constraint on.
+
+        `distribution` is a dictionary where keys represent the unique values and the corresponding
+        tuple values represent the minimum and maximum allowed proportions of the respective
+        unique value in the columns.
+
+        `default_bounds` A tuple specifying the minimum and maximum allowed proportions for all
+        elements not mentioned in the distribution. By default, it's set to (0, 0), which means
+        all elements not present in distribution will cause a constraint failure.
```
kklein, line 372:
```suggestion
        all elements not present `distribution` will cause a constraint failure.
```

src/datajudge/constraints/uniques.py
```diff
@@ -292,3 +294,76 @@ def compare(
...
+    """
+    VariantDistributionConstraint is a class to check if the distribution of values in a column
+    falls within the specified minimum and maximum bounds.
+    The constraint compares the factual distribution of values in a column of a `DataSource`
+    with the target distribution supplied as a dictionary where keys represent the unique values
```
kklein, line 304: Nitpick: I think the keys represent some unique values, not necessarily all of them, right?

src/datajudge/constraints/uniques.py
kklein: I agree.
What about `CategoricalBoundConstraint`? It doesn't roll beautifully off the tongue, either, though. ;)

If I understand correctly, we don't need to always specify an entire distribution. E.g. if there are distinct values A, B, C, D, E and F, we could also use this constraint - and not just for the sake of the argument, but with real value - to bound the shares of A, B, C, D - which together we assume to cover 99% of the rows. We could then set `default_bound=(0, .01)`.

src/datajudge/constraints/uniques.py
```diff
@@ -292,3 +294,76 @@ def compare(
...
+
+    Example use cases include testing for consistency in columns with expected categorical values
+    or ensuring that the distribution of values in a column adheres to a certain criterion.
+    """
+
+    def __init__(
+        self,
+        ref: DataReference,
+        distribution: Dict[T, Tuple[float, float]],
+        default_bounds: Tuple[float, float] = (0, 0),
```
kklein, line 320: What would you think of a relative violation tolerance parameter? E.g. it could say:

A test succeeds iff
#observations outside of the specified ranges / #observations <= tolerance_parameter

I don't consider it a must - we've simply faren well with tolerances historically.

src/datajudge/constraints/uniques.py
```diff
@@ -292,3 +294,76 @@ def compare(
...
+        total = factual.total()
+        all_variants = factual.keys() | target.keys()
+        min_counts = Counter(
+            {k: target.get(k, self.default_bounds)[0] * total for k in all_variants}
+        )
+        max_counts = Counter(
+            {k: target.get(k, self.default_bounds)[1] * total for k in all_variants}
+        )
+
+        violations = (factual - max_counts) | (min_counts - factual)
```
kklein, line 346: Slick :)

kklein (commented):
Really great work!

src/datajudge/constraints/uniques.py
0xbe7a: If 'A' is expected to have a target share ranging from 5% to 15%, but its actual share is 16%, would you consider the 16% to be a violation of the target range or merely 1% above the upper limit?

tests/integration/test_integration.py
```diff
@@ -742,6 +742,41 @@ def test_n_uniques_loss_between(engine, unique_table2, unique_table1, data):
...
+            },
+            (0, 0),
+        ),
+        (identity, ["col_varchar"], {"hi0": (0.65, 0.85), "hi1": (0.1, 0.35)}, (0, 0)),
+        (negation, ["col_varchar"], {"hi0": (0.65, 0.85)}, (0, 0)),
+        (identity, ["col_varchar"], {"hi0": (0.65, 0.85)}, (0, 0.35)),
+        (negation, ["col_varchar"], {}, (0, 0.35)),
+    ],
+)
+def test_distribution_within(engine, distribution_table, data):
```
kklein, line 772: We might want to reflect the new constraint name in the test function name, e.g. `test_categorical_bound_within`.

src/datajudge/constraints/uniques.py
0xbe7a: I added the feature

PR #145, opened 2023-05-16
allow to specify sqlalchemy types directly in ColumnType Constraints

0xbe7a:
This PR introduces modifications to the ColumnType Constraint class and the add_column_type_constraint method to optionally allow users to directly pass SQLAlchemy Types.

The aim is to improve user experience by leveraging SQLAlchemy's type hierarchy and mitigating backend compatibility issues.

Consider a scenario where a user wants to ensure a column, 'price', in their database should always store numeric values. However, due to the variability in different databases, this numeric value could be stored as INTEGER, DECIMAL, DOUBLE, or other numeric types depending on the backend, which could change over time.
With the changes in this PR, the user can now use SQLAlchemy's generic Numeric type, which matches any numeric type in the backend.

All changes have been tested with existing integration tests and are open for further feedback and testing.

### DIFF
--- a/src/datajudge/constraints/column.py
+++ b/src/datajudge/constraints/column.py
@@ -1,5 +1,5 @@
 import abc
-from typing import List, Tuple
+from typing import List, Optional, Tuple, Union
 
 import sqlalchemy as sa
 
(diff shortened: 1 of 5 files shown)


src/datajudge/requirements.py
```diff
@@ -166,8 +176,30 @@ def add_uniqueness_constraint(
         )
 
     def add_column_type_constraint(
-        self, column: str, column_type: str, name: str = None
+        self,
+        column: str,
+        column_type: Union[str, sa.types.TypeEngine],
```
ivergara, line 181: Why don't you make a type alias to this?

0xbe7a: I'm not sure if it would help with readability and understandability, there are only a handful of usages and it adds a layer of indirection needed when reading the documentation.

src/datajudge/constraints/column.py
```diff
@@ -69,29 +69,48 @@ def compare(
...
+        if isinstance(column_type_target, sa.types.TypeEngine):
+            result = isinstance(column_type_factual, type(column_type_target))
+        else:
+            column_type = str(column_type_factual).lower()
+            # Integer columns loaded from snowflake database may be referred to as decimal with
+            # 0 scale. More here:
+            # https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#decimal-numeric
+            if column_type == "decimal(38, 0)":
+                column_type = "integer"
+            result = column_type.startswith(column_type_target.lower())
```
kklein, line 115: That's much nicer than the splitting we did before. :)

src/datajudge/constraints/column.py
```diff
@@ -69,29 +69,48 @@ def compare(
...
-        result = column_type_factual.startswith(column_type_target)
+
+        if isinstance(column_type_target, sa.types.TypeEngine):
+            result = isinstance(column_type_factual, type(column_type_target))
+        else:
+            column_type = str(column_type_factual).lower()
+            # Integer columns loaded from snowflake database may be referred to as decimal with
+            # 0 scale. More here:
+            # https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#decimal-numeric
+            if column_type == "decimal(38, 0)":
```
kklein, line 113: Afaict we previously handled this case for snowflake only. Did you extend it by first principles or solely because it didn't make any integration tests fail?

kklein (approved):
Lgtm! Thanks!

src/datajudge/constraints/column.py
0xbe7a: Didn't make any integration tests fail and I wanted to avoid tracking which concrete DB Engine the column type was obtained from and maintaining a custom mapping which is dependent on the DB engine

ivergara (approved):
Good job!

Issue #152, opened 2023-06-06
Improve error message if str was passed where list[str] was expected


jonashaag:
If you pass a `str` to `add_categorical_bound_constraint(columns=)` it will error with `KeyError: 'x'` where *x* is the first letter of the column name. Maybe we can have it output a more helpful error message.

ivergara:
What we can do is that if the argument is not a Collection then it fails with a better error message.

PR #155, opened 2023-06-09
Provide better errors messages, if user passes a single column name instead of a list of columns

0xbe7a:
Currently, if a user specifies only one column name instead of a list of column names, the resulting error messages may occur late in execution and be opaque (see #152). This PR adds better error messages for this case.

### DIFF
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -6,7 +6,7 @@ import operator
 from abc import ABC, abstractmethod
 from collections import Counter
 from dataclasses import dataclass
-from typing import Callable, Optional, Sequence, final, overload
+from typing import Callable, Sequence, final, overload
 
 import sqlalchemy as sa
 from sqlalchemy.sql import selectable
@@ -294,8 +294,8 @@ class DataReference:
     def __init__(
         self,
         data_source: DataSource,
-        columns: Optional[list[str]] = None,
-        condition: Optional[Condition] = None,
+        columns: list[str] | None = None,
+        condition: Condition | None = None,
     ):
         if columns is not None and not isinstance(columns, list):
             raise TypeError(f"columns must be a list, not {type(columns)}")
--- a/src/datajudge/db_access.py
+++ b/src/datajudge/db_access.py
@@ -6,7 +6,7 @@ import operator
 from abc import ABC, abstractmethod
 from collections import Counter
 from dataclasses import dataclass
-from typing import Callable, Sequence, final, overload
+from typing import Callable, Optional, Sequence, final, overload
 
 import sqlalchemy as sa
 from sqlalchemy.sql import selectable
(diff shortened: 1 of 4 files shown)


src/datajudge/db_access.py
```diff
@@ -294,9 +294,12 @@ class DataReference:
     def __init__(
         self,
         data_source: DataSource,
-        columns: list[str] = None,
-        condition: Condition = None,
+        columns: Optional[list[str]] = None,
```
ivergara, line 297: @kklein are you aware if `columns` needs to be a list, aka, is the order of the elements relevant?

src/datajudge/db_access.py
```diff
@@ -294,9 +294,12 @@ class DataReference:
...
+        condition: Optional[Condition] = None,
     ):
+        if columns is not None and not isinstance(columns, list):
```
ivergara, line 300: I think we should generalize a bit the `isinstance` call depending on the answer in [here](https://github.com/Quantco/datajudge/pull/155/files#r1223975407),

If the answer is yes, then here probably we should compare to something like `Sequence`. If the answer is no, then we should compare to `Collection`.

I do realize that there will be a slight mismatch between the annotation in the function signature and what's happening inside at runtime, but we could fix the annotation in a second step.

src/datajudge/db_access.py
kklein: I think we have some cases where we select the first element of a list (e.g. as a 'sample') with `columns[0]`. Besides this implementation detail I don't think we need ordering and I don't think we want repetition.

kklein: E.g. here https://github.com/Quantco/datajudge/blob/main/src/datajudge/constraints/date.py#L171

ivergara: But that access is more like "hey, there should be only one".

Then, I'd suggest to consider changing the annotation to `Collection[str]` following the mantra 

> accept interfaces, return concrete types

kklein: I'd much prefer that, too - I just wanted to chime in that some internal bits and pieces might need to be adapted; test would need to be adapted, too, to figure out which parts need adaptation).

0xbe7a: As a test, I randomly shuffled the Column Names in the constructor: only the Unique Constraints break.

The `get_uniques` in db_access.py constructs tuples whose order depends on the order in DataReference.
https://github.com/Quantco/datajudge/blob/main/src/datajudge/db_access.py#LL913C1-L913C1

The Uniques Between Constraints construct the list of tuples for both tables and then compare both lists. 
 For this, their DataReference must also have the same order, otherwise the tuples with values cannot be compared. Alternatively we could construct a dict

ivergara: Then, instead of `Collection` we for for `Sequence` for now.

kklein (approved):

kklein:
I think a changelog entry could be useful!

src/datajudge/db_access.py
ivergara: Fair point. Then let's tackle this in a separate discussion.

ivergara (approved):

PR #163, opened 2023-06-16
Update ci.yaml

ivergara:
Allowing fork-pull workflow

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,7 +80,6 @@ jobs:
       - name: Checkout branch
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.head_ref }}
           fetch-depth: 0
       - name: Set up micromamba
         uses: mamba-org/setup-micromamba@875557da4ee020f18df03b8910a42203fbf02da1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -42,8 +42,6 @@ jobs:
     steps:
       - name: Checkout branch
         uses: actions/checkout@v3
-        with:
```
kklein, line 45: Would we also need to remove this for all of the respective integration tests?

pavelzw: I'd say so

ivergara: Good catch!

.github/workflows/ci.yaml
```diff
@@ -82,7 +80,6 @@ jobs:
       - name: Checkout branch
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.head_ref }}
           fetch-depth: 0
```
AliPiccioniQC, line 83: nit: This fetches the entire history. Is this to run some backwards compatibility test?

AliPiccioniQC (approved):

.github/workflows/ci.yaml
ivergara: You mean it should be `fetch-depth: 1`? Which most likely is the default I guess.

I don't think there is any good reason why we do this.

AliPiccioniQC: Yeah might be better to just remove the entire with block for simplicity.

pavelzw: I think `fetch-depth` is used in our repositories because `setuptools-scm` uses Git tags to determine the version when building.

kklein (approved):
Thanks a lot for looking into this!

PR #166, opened 2023-06-18
Refactor: `DataReference`: Use `__str__`  instead of `get_string`

kklein:
I struggle to reconstruct why we originally didn't opt for this. I figure the `get_string` method used to rely on an additional argument it no longer depends on - just a guess, though.

### DIFF
--- a/src/datajudge/constraints/base.py
+++ b/src/datajudge/constraints/base.py
@@ -149,7 +149,7 @@ class Constraint(abc.ABC):
 
     @property
     def target_prefix(self) -> str:
-        return f"{str(self.ref2)}'s " if (self.ref2 is not None) else ""
+        return f"{self.ref2}'s " if (self.ref2 is not None) else ""
 
     @property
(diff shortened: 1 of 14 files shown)


src/datajudge/constraints/base.py
```diff
@@ -149,7 +149,7 @@ def get_description(self) -> str:
 
     @property
     def target_prefix(self) -> str:
-        return f"{self.ref2.get_string()}'s " if (self.ref2 is not None) else ""
+        return f"{str(self.ref2)}'s " if (self.ref2 is not None) else ""
```
ivergara, line 152: It should be sufficient to do as follows
```suggestion
        return f"{self.ref2}'s " if (self.ref2 is not None) else ""
```

The `f-string` already calls the `__str__` method of the class.

ivergara (approved):
Looks good!

kklein:
@ivergara Great catch :)

ivergara (approved):

PR #170, opened 2023-06-22
Add formatting for assertion messages

0xbe7a:
This PR adds styling codes for assertion messages.
This allows constraints to use styling e.g. to increase readability of their assertion messages. The styling can then be set independently of the platform and converted to ANSI color codes for command line output, for example, and to CSS color tags for HTML reports. The styling tags describe use cases and not concrete colors, so formatter can use arbitrary color palettes and these are not fixed by the constraint. The first use-case which is supported is: numMatch and numDiff for (#89)

The assertion message can be formatted as follows: `[numMatch]3.141[/numMatch][numDifference]6[/numDifference] vs [numMatch]3.141[/numMatch][numDifference]5[/numDifference]`

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,6 @@ dependencies:
   - python>=3.8
   - pytest
   - pytest-cov
-  - conda-forge/label/pytest-html_rc::pytest-html==4.0.0rc4
   - pytest-xdist
   - make
   - numpydoc
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ dependencies:
   - python>=3.8
   - pytest
   - pytest-cov
-  - pytest-html
+  - conda-forge/label/pytest-html_rc::pytest-html==4.0.0rc4
   - pytest-xdist
   - make
   - numpydoc
--- a/src/datajudge/constraints/base.py
+++ b/src/datajudge/constraints/base.py
@@ -6,6 +6,9 @@ from typing import Any, Callable, List, Optional, Tuple, TypeVar
 import sqlalchemy as sa
 
 from ..db_access import DataReference
+from ..formatter import DefaultFormatter, Formatter
+
+DEFAULT_FORMATTER = DefaultFormatter()
 
 T = TypeVar("T")
 OptionalSelections = Optional[List[sa.sql.expression.Select]]
@@ -25,10 +28,30 @@ def uncommon_substrings(string1: str, string2: str) -> Tuple[str, str]:
 @dataclass(frozen=True)
 class TestResult:
     outcome: bool
-    failure_message: Optional[str] = field(default=None, repr=False)
-    constraint_description: Optional[str] = field(default=None, repr=False)
+    _failure_message: Optional[str] = field(default=None, repr=False)
+    _constraint_description: Optional[str] = field(default=None, repr=False)
     _factual_queries: Optional[str] = field(default=None, repr=False)
     _target_queries: Optional[str] = field(default=None, repr=False)
+
+    def formatted_failure_message(self, formatter: Formatter) -> Optional[str]:
+        return (
+            formatter.fmt_str(self._failure_message) if self._failure_message else None
+        )
+
+    def formatted_constraint_description(self, formatter: Formatter) -> Optional[str]:
+        return (
+            formatter.fmt_str(self._constraint_description)
+            if self._constraint_description
+            else None
+        )
+
+    @property
+    def failure_message(self) -> Optional[str]:
+        return self.formatted_failure_message(DEFAULT_FORMATTER)
+
+    @property
+    def constraint_description(self) -> Optional[str]:
+        return self.formatted_constraint_description(DEFAULT_FORMATTER)
 
     @property
     def logging_message(self):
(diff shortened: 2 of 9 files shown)


src/datajudge/constraints/base.py
```diff
@@ -25,11 +28,31 @@ def uncommon_substrings(string1: str, string2: str) -> Tuple[str, str]:
...
+
+    def formatted_constraint_description(self, formatter: Formatter) -> Optional[str]:
+        return (
+            formatter.fmt_str(self._constraint_description)
+            if self._constraint_description
+            else None
+        )
+
+    @property
+    def failure_message(self) -> Optional[str]:
```
ivergara, line 49: I like how you are using properties as one of the cookbook intentions, to add behavior to an instance attribute.

src/datajudge/formatter.py
```diff
@@ -0,0 +1,50 @@
...
+STYLING_CODES = r"\[(numMatch|numDiff)\](.*?)\[/\1\]"
+
+
+class Formatter(abc.ABC):
+    @abc.abstractmethod
+    def fmt_str(self, result: str) -> str:
+        pass
+
+
+class DefaultFormatter(Formatter):
```
ivergara, line 15: Mostly stylistic, but I'd put the implementation of the `__init__` and `fmt_str` directly in the base class and then let all specific implementations derive from that one instead of the `DefaultFormatter`.

src/datajudge/formatter.py
```diff
@@ -0,0 +1,50 @@
...
+        # Replace codes with platform specific styling
+        string = self.known_bb_pattern.sub(
+            lambda m: self.apply_formatting(m.group(1), m.group(2)), string
+        )
+
+        return string
+
+
+class AnsiColorFormatter(DefaultFormatter):
+    def __init__(self):
```
ivergara, line 33: I don't think you need this here. At least you're not using it in `HtmlFormatter`.

ivergara (commented):
Would be great if you add something to the PR body, and also a paragraph somewhere in the documentation about this feature.

Probably also add an entry to the changelog.

YYYasin19 (approved):
Nice!
~Can you add a little paragraph in the documentation? That way we have a constant reminder that we'll need to extend this in the future :)~

src/datajudge/pytest_integration.py
```diff
@@ -4,9 +4,23 @@
...
+def formatter(pytestconfig):
+    color = pytestconfig.getoption("color")
+    is_html = pytestconfig.getoption("htmlpath") is not None
+
+    if not is_html and (color == "yes" or color == "auto"):
+        return AnsiColorFormatter()
+    elif is_html and (color == "yes" or color == "auto"):
+        return HtmlFormatter()
+    else:
+        return Formatter()
```
YYYasin19, line 21: Could you add some tests that validate that this works?
e.g. something that creates a temp test file and calls `pytest` with different options on it.

0xbe7a:
While i was trying to get html formatting to work, i noticed that there is a new major release 4.0 for pytest coming up. When ansi2html was installed together with pytest-html (the package was not a dependency due to the more permissive software license of ansi2html) pytest-html could already translate the ansi styling to css styling in the caputred log output. Styling in assertion messages was however afaik not supported / bugged. With the rewrite of pytest-html this problem seems to be fixed and pytest-html can now also translate styling in assertion messages itself. This makes at least the html formatter obsolete for us, but I would argue that a separate styling backend is still useful to provide the option for unstyled output for environments without ansi coloring support. I am currently in the process of getting the current 4.0.0 release canidate of pytest-html into conda-forge to allow for testing.

ivergara:
> I am currently in the process of getting the current 4.0.0 release canidate of pytest-html into conda-forge to allow for testing.

You don't have to. Just install from PyPi into your local environment.

0xbe7a:
> You don't have to. Just install from PyPi into your local environment.

yes, for local testing this works of course. but to work on this pr and test if our integration tests etc. work as well it is handy to be able to target the rc directly in the environment.yml so it can be used directly in the the ci tests too

src/datajudge/pytest_integration.py
0xbe7a: I have added an integration test to check if the formatter is selected correctly

environment.yml
```diff
@@ -6,7 +6,7 @@ dependencies:
   - python>=3.8
   - pytest
   - pytest-cov
-  - pytest-html
+  - conda-forge/label/pytest-html_rc::pytest-html==4.0.0rc4
```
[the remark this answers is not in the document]
0xbe7a: Yes, i think we should wait for the proper release before merging this PR, i just wanted to get it ready for review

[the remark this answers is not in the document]
0xbe7a: I like the approach of soft support, but not hard dependence, will add a commit!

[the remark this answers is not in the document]
ivergara: > Yes, i think we should wait for the proper release before merging this PR, i just wanted to get it ready for review

That'd be fine if we know the release will be done soonish.

In any case, we should support, as in not break, if a user has a pytest-html < 4 version installed. In that case the html then doesn't have coloring but should break the output there.

ivergara (approved):

PR #224, opened 2024-06-03
adding configuration options to uniques functionality

SimonLangerQC:


src/datajudge/constraints/uniques.py
```diff
@@ -63,6 +74,31 @@ class Uniques(Constraint, abc.ABC):
...
+
+    For `UniquesSubset`, by default,
+    the number of occurrences affects the computed fraction of violations.
+    To disable this weighting, set `compare_distinct=True`.
+    This argument does not have an effect on the test results for other `Uniques` constraints,
+    or if `max_relative_violations` is 0.
+
+    By default, the assertion messages make use of sets,
+    thus, they may differ from run to run despite the exact same situation being present.
+    To enforce a reproducible output via (e.g.) sorting, set `output_postprocessing_sorter` to a callable
```
ivergara, line 90: I think that for now we can simplify and not add an arbitrary posprocessing step and just make the messages reproducible.

src/datajudge/constraints/uniques.py
```diff
@@ -63,6 +74,31 @@ class Uniques(Constraint, abc.ABC):
...
+    which takes in two collections, and returns modified (e.g. sorted) versions of them.
+    In most cases, the second argument is simply None,
+    but for `UniquesSubset` it is the counts of each of the elements.
+    The suggested function is ``datajudge.utils.util_output_postprocessing_sorter`` from this file,
+    - see its documentation for details.
+
+    By default, the number of subset or superset remainders (excess or missing values)
```
ivergara, line 97: Maybe we can improve this by using pytest's `-vv` mode somehow? Perhaps to be dealt in a separate PR?

src/datajudge/utils.py
```diff
@@ -42,3 +44,106 @@ def format_difference(
         f"{s1[:diff_idx]}{_fmt_diff_part(s1, diff_idx)}",
         f"{s2[:diff_idx]}{_fmt_diff_part(s2, diff_idx)}",
     )
+
+
+def util_output_postprocessing_sorter(
```
ivergara, line 49: I was goign to suggest a move like this. Now, since these new functions are already in a `utils` namespace, doesn't make much sense to prepend such set of characters for each function.

src/datajudge/utils.py
```diff
@@ -42,3 +44,106 @@ def format_difference(
...
+    if len(collection) <= 1:
+        return collection, counts  # empty or 1 element lists are always sorted
+
+    lst = sort_tuple_none_aware(
+        [(-count, *elem) for count, elem in zip(counts, collection)]
+    )
+    return [elem[1:] for elem in lst], [-elem[0] for elem in lst]
+
+
+def util_filternull_default_deprecated(values: List[T]) -> List[T]:
```
ivergara, line 85: This is the default behavior until now, which you're using when the `None` is used. But eventually still makes sense to use it in the future right? If so, I don't think that the function itself is deprecated and having that word in the function name is misleading.

tests/integration/test_integration.py
```diff
@@ -535,6 +801,608 @@ def test_uniques_subset_within(engine, unique_table1, data):
...
+            "(1, 'hi1'), (4, 'hi2'), (4, 'hi3'), (7, 'hi4'), (7, 'hi5'), (10, 'hi6'), (10, 'hi7'), (13, "
+            "'hi8'), (13, 'hi9'), (16, 'hi10'), (16, 'hi11'), (19, 'hi12'), (19, 'hi13'), (22, 'hi14'), (22, "
+            "'hi15'), (25, 'hi16'), (25, 'hi17'), (28, 'hi18'), (28, 'hi19')]' with counts [2, 2, 2, 2, 2, "
+            "2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, "
+            "1, 1, 1].",
+        ),
+        (
+            negation,
+            ["col_int"],
+            [
```
ivergara, line 849: Maybe we instruct our linter/formatter not to do one element per line in this list? Is really unnecessarily bloating the line count. Same for the other parameterizations.

Alternatively, perhaps you can do something like `list(range(30))` instead.

ivergara (commented):
Quite some lines! Greatly appreciated.

1. As written in comments, perhaps you can leave the postprocessing parts of this PR out to keep its size more manageable and to have a better discussion on what to do there. I'm not strongly against dealing with the postprocessing here.

2. The names of the utility functions feel a bit wonky to me. But right now I dont'know how to name them better.

3. In the utility functions, you use `filter`, `reduce` with `lambda` functions calling `list` at the end. Is there any chance to change some of them to list comprehensions instead?

src/datajudge/constraints/uniques.py
```diff
@@ -42,9 +44,18 @@ class Uniques(Constraint, abc.ABC):
...
 
-    Null values in the column are ignored. To assert the non-existence of them use
+    Null values in the column are ignored by default. To assert the non-existence of them use
     the `NullAbsence` constraint via the `add_null_absence_constraint` helper method for
     `WithinRequirement`.
+    By default, the null filtering does not trigger if multiple columns are fetched at once.
+    It can be configured in more detail by supplying a custom ``filter_func`` function.
+    Some exemplary implementations are available in this module as ``datajudge.utils.util_filternull_default_deprecated``,
+    ``datajudge.utils.util_filternull_never``, ``datajudge.utils.util_filternull_element_or_tuple_all``, ``datajudge.utils.util_filternull_element_or_tuple_any``.
+    For new deployments, using one of the above filters or a custom one is recommended.
```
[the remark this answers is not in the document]
SimonLangerQC: Removed.

src/datajudge/constraints/uniques.py
SimonLangerQC: The main reason why I wanted to make this postprocessing step so generic is that there may be some users with column datatypes which are not (automatically) sortable. If we always force the output_postprocessing_sorter to my suggested function, existing code would break, with no easy way to fix/turn off that feature.

I initially also considered just adding a binary sorted_output flag, but it seemed less flexible:
1. the above mentioned case, requiring intervention to make the output sortable
2. to give flexibility depending w.r.t ASC/DESC and how to order NULL values, which is quite use-case dependent if they are interesting, or should rather be at the end of the output (possibly cut off)

src/datajudge/constraints/uniques.py
SimonLangerQC: I did not do this for 2 reasons:
1. Datajudge currently does not depend on pytest, introducing the need for some external configuration option anyway.
https://github.com/Quantco/datajudge/blob/efc54e3d52a5288abfbf4afe4098bf3c0b0c3ae6/pyproject.toml#L27-L32

2. Whether the user wants much or little output could differ from one Constraints to the next --> I wanted to give granular control.
3. Our `_with_outputcheck` tests fail if a different pytest configuration is used

src/datajudge/utils.py
SimonLangerQC: My main issue with the current default behavior is that if the user adds a second column to the unique constraint, the NULL-filtering no longer works, even if both the original and second column are NULL.

I've renamed it to `filternull_element`

SimonLangerQC:
Thanks for the thorough review @ivergara :) - I've integrated your suggestions or replied to your comments.

Factoring out the postprocessing imo is not a great idea, since this would break the `_with_outputcheck` tests, which are also used to verify the updated null-filtering, etc.

SimonLangerQC:
I've now moved the additional output configuration options to the base `Constraint` class, allowing for making use of sorting and slicing for other types of constraints as well, such as the functional dependency constraint

run_integration_tests_postgres.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+docker stop $(docker ps -q --filter name=postgres_datajudge)
+
+./start_postgres.sh &
+bash -c "while true; do printf '\nPress enter to once postgres is ready: '; sleep 1; done" &
```
ivergara, line 6:
```suggestion
bash -c "while true; do printf '\nPress enter once postgres is ready: '; sleep 1; done" & 
```
?

src/datajudge/constraints/base.py
```diff
@@ -113,7 +113,16 @@ class Constraint(abc.ABC):
...
 
     def __init__(
-        self, ref: DataReference, *, ref2=None, ref_value: Any = None, name: str = None
+        self,
+        ref: DataReference,
+        *,
+        ref2=None,
+        ref_value: Any = None,
+        name: str = None,
+        output_postprocessing_sorter: Callable[
```
ivergara, line 122: Is this only capable of doing sorting? Probably you could do any kind of postprocessing of the output. As such, I'd suggest changing the name of the argument to `output_processor`or if you want to be more precise `output_postprocessor`.

src/datajudge/constraints/base.py
```diff
@@ -113,7 +113,16 @@ class Constraint(abc.ABC):
...
+            [Collection, Optional[Collection]], Collection
+        ] = None,
+        output_remainder_slicer=slice(5),
```
ivergara, line 125: Could the logic for this "slicer" be given as a postprocessor instead of needing an extra argument?

SimonLangerQC: Not easily, since the postprocessor is applied in more places than the slicing, there are many instances where the slicing was not present before in the library, and i did not want to introduce it there

tests/integration/test_integration.py
```diff
@@ -535,6 +806,308 @@ def test_uniques_subset_within(engine, unique_table1, data):
...
+            output_postprocessing_sorter,
+            slice(None),
+            None,
+            None,
+            "column(s) 'col_int' has a fraction of 0.03333333333333333 > 0 values (2 / 60) not being an element of '[0, 1, 2, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29]'. It has e.g. (slice(None, None, None)) excess elements '[3]' with counts [2].",
+        ),
+        (
+            negation,
+            ["col_int"],
+            [i for i in range(30) if i != 3],
```
ivergara, line 906: Another reason why this is better, now I explicitly see that the number 3 was missing from the previous list!

src/datajudge/constraints/base.py
SimonLangerQC: Done

src/datajudge/constraints/base.py
```diff
@@ -119,10 +120,7 @@ def __init__(
         ref2=None,
         ref_value: Any = None,
         name: str = None,
-        output_processor: Callable[
-            [Collection, Optional[Collection]], Collection
-        ] = None,
-        output_remainder_slicer=slice(5),
+        output_processors: List[OutputProcessor] = None,
```
ivergara, line 123: As discussed you could define this as `OutputProcessor | List[OutputProcessor]` and in the body you can do something like

```python
if not isinstance(output_processor, List]:
    output_processor = [output_processor]
```

SimonLangerQC: Done

src/datajudge/requirements.py
```diff
@@ -1420,8 +1485,19 @@ def add_uniques_equality_constraint(
...
         columns.
 
-        See the ``Uniques`` class for further parameter details on ``map_func`` and
-        ``reduce_func``.
+        Null values in the column are ignored. To assert the non-existence of them use
+        the ``NullAbsence`` constraint via the ``add_null_absence_constraint`` helper method
+        for ``WithinRequirement``.
+        By default, the null filtering does not trigger if multiple columns are fetched at once.
+        It can be configured in more detail by supplying a custom ``filter_func`` function.
+        Some exemplary implementations are available in this module as ``datajudge.utils.filternull_element``,
```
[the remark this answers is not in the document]
kklein: If, in addition, we provide a ~-prefix, the the function is no longer qualified:

Screenshot 2024-06-06 at 21 00 15

src/datajudge/requirements.py
```diff
@@ -1464,7 +1553,8 @@ def add_uniques_superset_constraint(
         One use of this constraint is to test for consistency in columns with expected
         categorical values.
 
-        See ``Uniques`` for further details on ``map_func`` and ``reduce_func``.
+        See ``Uniques`` for further details on ``map_func``, ``reduce_func``,
```
kklein, line 1556:
```suggestion
        See :class:`~datajudge.constraints.uniques.Uniques` for further details on ``map_func``, ``reduce_func``,
```

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
         Null values in the column are ignored. To assert the non-existence of them use
         the ``NullAbsence`` constraint via the ``add_null_absence_constraint`` helper method
         for ``WithinRequirement``.
+        By default, the null filtering does not trigger if multiple columns are fetched at once.
+        It can be configured in more detail by supplying a custom ``filter_func`` function.
+        Some exemplary implementations are available in this module as ``datajudge.utils.filternull_element``,
```
kklein, line 1602:
```suggestion
        Some exemplary implementations are available in this module as :func:`~datajudge.utils.filternull_element`,
```

kklein: Same question as before - is it really 'this' module?

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
...
+        ``datajudge.utils.filternull_never``, ``datajudge.utils.filternull_element_or_tuple_all``, ``datajudge.utils.filternull_element_or_tuple_any``.
```
kklein, line 1603:
```suggestion
        :func:`~datajudge.utils.filternull_never`, :func:`~datajudge.utils.filternull_element_or_tuple_all`, :func:`~datajudge.utils.filternull_element_or_tuple_any`.
```

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
...
+        Passing None as the argument is equivalent to ``datajudge.utils.filternull_element``, but triggers a warning.
```
kklein, line 1604:
```suggestion
        Passing ``None`` as the argument is equivalent to :func:`~datajudge.utils.filternull_element` but triggers a warning.
```

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
...
+        The deprecated default may change in future versions.
+        To silence the warning, set ``filter_func`` explicitly.
 
         ``max_relative_violations`` indicates what fraction of rows of the given table
         may have values not included in the reference set of unique values. Please note
         that ``UniquesSubset`` and ``UniquesSuperset`` are not symmetrical in this regard.
 
-        See ``Uniques`` for further details on ``map_func`` and ``reduce_func``.
+        By default, the number of occurrences affects the computed fraction of violations.
+        To disable this weighting, set `compare_distinct=True`.
```
kklein, line 1613:
```suggestion
        To disable this weighting, set ``compare_distinct=True``.
```

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
...
+        This argument does not have an effect on the test results for other `Uniques` constraints,
+        or if `max_relative_violations` is 0.
```
kklein, line 1615:
```suggestion
        or if ``max_relative_violations`` is 0.
```

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
...
+        To silence the warning, set ``filter_func`` explicitly.
 
         ``max_relative_violations`` indicates what fraction of rows of the given table
         may have values not included in the reference set of unique values. Please note
         that ``UniquesSubset`` and ``UniquesSuperset`` are not symmetrical in this regard.
 
-        See ``Uniques`` for further details on ``map_func`` and ``reduce_func``.
+        By default, the number of occurrences affects the computed fraction of violations.
+        To disable this weighting, set `compare_distinct=True`.
+        This argument does not have an effect on the test results for other `Uniques` constraints,
```
kklein, line 1614:
```suggestion
        This argument does not have an effect on the test results for other :class:`~datajudge.constraints.uniques.Uniques` constraints,
```

src/datajudge/requirements.py
```diff
@@ -1500,12 +1597,25 @@ def add_uniques_subset_constraint(
...
+        or if `max_relative_violations` is 0.
+
+        See ``Uniques`` for further details on ``map_func``, ``reduce_func``,
```
kklein, line 1617:
```suggestion
        See :class:`~datajudge.constraints.uniques.Uniques` for further details on ``map_func``, ``reduce_func``,
```

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+) -> Collection:
+    """
+    Sorts a collection of tuple elements in descending order of their counts,
+    and for ties, makes use of the ascending order of the elements themselves.
+
+    If the first element is not instanceof tuple,
+    each element will be transparently packaged into a 1-tuple for processing;
+    this process is not visible to the caller.
+
+    Handles None values as described in `sort_tuple_none_aware`.
```
kklein, line 66:
```suggestion
    Handles ``None`` values as described in ``sort_tuple_none_aware``.
```

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+def filternull_element_or_tuple_any(values: List) -> List:
+    return [
+        value
+        for value in values
+        if value is not None
+        and not (isinstance(value, tuple) and any(x is None for x in value))
+    ]
+
+
+def sort_tuple_none_aware(collection: Collection[Tuple], ascending=True):
```
kklein, line 143: I think it could be useful to add a type hint for the return value here. :)

src/datajudge/requirements.py
```diff
@@ -270,17 +275,30 @@ def add_uniques_equality_constraint(
         of a ``DataSource`` are strictly the ones of a reference set of expected values,
         specified via the ``uniques`` parameter.
 
+        Null values in the column are ignored. To assert the non-existence of them use
```
kklein, line 278:
```suggestion
        Null values in the columns ``colums`` are ignored. To assert the non-existence of them use
```

src/datajudge/requirements.py
```diff
@@ -270,17 +275,30 @@ def add_uniques_equality_constraint(
...
+        the ``NullAbsence`` constraint via the ``add_null_absence_constraint`` helper method
```
kklein, line 279: Afaict there isn't really a `NullAbsence` constraint but `add_null_absence_constraint` creates a `MaxNullFraction constraint.

```suggestion
        the :meth:`~datajudge.requirements.WithinRequirement.add_null_absence_constraint`` helper method
```

src/datajudge/requirements.py
```diff
@@ -270,17 +275,30 @@ def add_uniques_equality_constraint(
...
+        for ``WithinRequirement``.
+        By default, the null filtering does not trigger if multiple columns are fetched at once.
+        It can be configured in more detail by supplying a custom ``filter_func`` function.
+        Some exemplary implementations are available in this module as ``datajudge.utils.filternull_element``,
```
kklein, line 283:
```suggestion
        Some exemplary implementations are available in this module as :func:`~datajudge.utils.filternull_element`,
```

src/datajudge/requirements.py
```diff
@@ -270,17 +275,30 @@ def add_uniques_equality_constraint(
...
+        ``datajudge.utils.filternull_never``, ``datajudge.utils.filternull_element_or_tuple_all``, ``datajudge.utils.filternull_element_or_tuple_any``.
```
kklein, line 284:
```suggestion
        :func:`~datajudge.utils.filternull_never`, :func:`~datajudge.utils.filternull_element_or_tuple_all`, :func:`~datajudge.utils.filternull_element_or_tuple_any`.
```

src/datajudge/requirements.py
```diff
@@ -270,17 +275,30 @@ def add_uniques_equality_constraint(
...
+        Passing None as the argument is equivalent to ``datajudge.utils.filternull_element``, but triggers a warning.
```
kklein, line 285:
```suggestion
        Passing ``None`` as the argument is equivalent to :func:`~datajudge.utils.filternull_element` but triggers a warning.
```

src/datajudge/requirements.py
```diff
@@ -270,17 +275,30 @@ def add_uniques_equality_constraint(
...
+        The deprecated default may change in future versions.
```
kklein, line 286: What is meant by that? That the deprecated setup will be removed or that a different default might be deprecated or that the default will become the non-deprecated option?

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+    if not isinstance(collection[0], tuple):
+        # package into a 1 tuple and pass into the method again
+        packaged_list = [(elem,) for elem in collection]
+        res_main, res_counts = output_processor_sort(packaged_list, counts)
+        return [elem[0] for elem in res_main], res_counts
+
+    if counts is None:
+        return sort_tuple_none_aware(collection), counts
+
+    assert len(collection) == len(
```
kklein, line 78: I would advocate against using assert for anything that is neither
* debugging nor
* testing.
Instead, I would suggest to explicitly raise Exceptions, in this case for instance a `ValueError`. 

The reason behind this is that `assert` is not actually guaranteed to be executed; it can be compiled away:
https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+        for value in values
+        if value is not None
+        and not (isinstance(value, tuple) and any(x is None for x in value))
+    ]
+
+
+def sort_tuple_none_aware(collection: Collection[Tuple], ascending=True):
+    """
+    Sorts a collection of either tuples or single elements,
+    where `None` is considered the same as the default value of the respective column's type.
```
kklein, line 146:
```suggestion
    where ``None`` is considered the same as the default value of the respective column's type.
```

What column are we referring to, here? Looking at this function in isolation it might not be too clear.

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+    For ints/floats `int()`/`float()` yield `0`/`0.0`, for strings `str()` yields `''`.
```
kklein, line 147:
```suggestion
    For ints and floats ``int()`` and ``float()`` yield ``0`` and ``0.0`` respectively; for strings, ``str()`` yields ``''``.
```

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+    The constructor is determined by calling type() on the first non-`None` element of the respective column.
```
kklein, line 148:
```suggestion
    The constructor is determined by calling ``type`` on the first non-``None`` element of the respective column.
```

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+
+    Checks and requires all elements in collection are tuples, and that all tuples have the same length.
```
kklein, line 150:
```suggestion
    Validates that all elements in collection are tuples and that all tuples have the same length.
```

src/datajudge/utils.py
```diff
@@ -42,3 +42,138 @@ def format_difference(
...
+    """
+    lst = list(collection)
+
+    if len(lst) <= 1:
+        return lst  # empty or 1 element lists are always sorted
+
+    assert all(
```
kklein, line 157: Same as above

kklein (commented):
Thanks a lot for your PR @SimonLangerQC - much appreciated!

I gave it a look; don't be surprised by the number of comments - most of them are merely docs-related polishing. :)

src/datajudge/requirements.py
```diff
@@ -1456,6 +1538,13 @@ def add_uniques_superset_constraint(
         Null values in the column are ignored. To assert the non-existence of them use
         the ``NullAbsence`` constraint via the ``add_null_absence_constraint`` helper method
         for ``WithinRequirement``.
+        By default, the null filtering does not trigger if multiple columns are fetched at once.
+        It can be configured in more detail by supplying a custom ``filter_func`` function.
+        Some exemplary implementations are available in this module as ``datajudge.utils.filternull_element``,
```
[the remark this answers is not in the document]
SimonLangerQC: Corrected.

src/datajudge/requirements.py
SimonLangerQC: Corrected.

src/datajudge/requirements.py
SimonLangerQC: Updated.

src/datajudge/utils.py
SimonLangerQC: Updated with more detailed explanation

src/datajudge/utils.py
SimonLangerQC: Done.

src/datajudge/requirements.py
SimonLangerQC: In-depth explanation added.

src/datajudge/utils.py
SimonLangerQC: Updated

SimonLangerQC:
Thanks for your comments @kklein - I've added your suggestions :)

src/datajudge/constraints/base.py
```diff
@@ -241,6 +256,12 @@ def test(self, engine: sa.engine.Engine) -> TestResult:
             target_queries,
         )
 
+    def apply_output_formatting_no_counts(self, values: Collection) -> Collection:
```
kklein, line 259:
```suggestion
    def apply_output_formatting(self, values: Collection) -> Collection:
```
Not quite sure what the `no_counts` is for here.

src/datajudge/utils.py
```diff
@@ -42,3 +42,143 @@ def format_difference(
...
+
+class OutputProcessor(Protocol):
+    def __call__(  # noqa: E704
+        self,
+        collection: Collection,
+        counts: Optional[Collection] = None,
+    ) -> Collection: ...
+
+
+def output_processor_sort(
```
kklein, line 55: Might it be useful to have unit tests for `output_processor_sort`, `output_processor_limit` and `sort_tuple_none_aware` in `datajudge/tests/unit/test_unit.py`?

src/datajudge/constraints/base.py
SimonLangerQC: This function does not support getting an additional counts variable to consider in its output formatting (relevant for the subset constraint - since that special case only occurs once (so far), the relevant code is directly integrated in it. But yeah, this makes the name unnecessarily long - updated.

src/datajudge/utils.py
SimonLangerQC: Done

kklein (approved):
LGTM! :)

PR #234, opened 2024-06-28
Allow disable cache

SimonLangerQC:
This PR adds a lru_cache_maxsize parameter to each constraint
--> user can control caching behavior and disable it in memory-constrained cases


ivergara (commented):
Why it seems that this PR does more than allowing disabling the cache?

src/datajudge/constraints/base.py
```diff
@@ -123,6 +127,7 @@ def __init__(
         output_processors: Optional[
             Union[OutputProcessor, List[OutputProcessor]]
         ] = output_processor_limit,
+        lru_cache_maxsize=None,
```
ivergara, line 130: I think using `lru_cache_maxsize` is a bit of a "mouthful". Would it work if it's `cache_size` instead? I'd say that the fact it's lru it's more of an implementation detail than something relevant to the user. You can tuck the `lru` and `max` parts away in the documentation for the curious.

SimonLangerQC:
> Why it seems that this PR does more than allowing disabling the cache?

There were considerable changes to the repo in between me creating the branch and now making the PR --> I had to resolve many merge conflicts. Looking into the files diff shows that I only add the new parameter (this happens in many locations: for each inheriting constraint, and for the `add_...` methods appropriate testing

src/datajudge/constraints/base.py
SimonLangerQC: Done with f14a54b

pixi.toml
```diff
@@ -88,6 +88,10 @@ sqlalchemy = "2.*"
 pytest-cov = "*"
 pytest-xdist = "*"
 
+[target.unix.dependencies]
```
ivergara, line 91: I guess you don't need `memray` in every unix-like installation, but only while developing perhaps.

As such, I'd suggest to create a dev feature/environment and define there a particular platform dependency

```toml
[feature.dev.target.unix.dependencies]
pytest-memray = "*"
memray = "*"
```

Even better, do this in a separate PR.

SimonLangerQC: Memray is used within a test case to verify that the caching is truly disabled --> required as a dependency.
Of course, if we do not want this additional overhead, we can simply remove the testcase: 

https://github.com/Quantco/datajudge/blob/f14a54b841ec535be1c70965d8cc873e42491aaa/tests/integration/test_integration.py#L1081-L1112

src/datajudge/constraints/base.py
```diff
@@ -156,13 +171,13 @@ def _check_if_valid_between_or_within(
                 f"{class_name}. Use exactly either of them."
             )
 
-    @lru_cache(maxsize=None)
+    # @lru_cache(maxsize=None)
```
0xbe7a, line 174:
```suggestion
    # @lru_cache(maxsize=None), see _setup_caching()
```

src/datajudge/constraints/base.py
```diff
@@ -156,13 +171,13 @@ def _check_if_valid_between_or_within(
...
     def get_factual_value(self, engine: sa.engine.Engine) -> Any:
         factual_value, factual_selections = self.retrieve(engine, self.ref)
         self.factual_selections = factual_selections
         return factual_value
 
-    @lru_cache(maxsize=None)
+    # @lru_cache(maxsize=None)
```
0xbe7a, line 180:
```suggestion
    # @lru_cache(maxsize=None), see _setup_caching()
```

tests/integration/test_integration.py
```diff
@@ -1040,6 +1061,109 @@ def test_uniques_subset_within_complex_with_outputcheck_extralong(
...
+        compare_distinct,
+        output_processors,
+        function,
+        condition,
+        failure_message_suffix,
+    ) = data
+
+    req = requirements.WithinRequirement.from_table(*unique_table_largesize)
+    for i in range(20):
+        req.add_uniques_subset_constraint(
```
0xbe7a, line 1098: Why do we need to add 20 constraints instead of calling `test()` 20 times?

tests/unit/test_utils.py
```diff
@@ -155,25 +155,29 @@ def test_output_processor_limit_defaults():
     input_counts = None
 
     input_main_copy = input_main.copy()
+    input_counts_copy = None
```
0xbe7a, line 158: ?

tests/integration/test_integration.py
SimonLangerQC: If we just called test 20x, it would re-use the cached data—no out-of-memory would be triggered. This test is meant to mimic a typical use case with many memory-intensive constraints being checked sequentially.

0xbe7a: Ah I see, please add a comment pointing out `cache_size=0` and the purpose of this test

tests/unit/test_utils.py
SimonLangerQC: I wanted to keep the test very similar to `test_output_processor_limit_withcounts`-testcase, where the counts are compared as well

tests/integration/test_integration.py
```diff
@@ -1040,6 +1061,109 @@ def test_uniques_subset_within_complex_with_outputcheck_extralong(
...
+    req.add_uniques_subset_constraint(
+        columns,
+        uniques,
+        max_relative_violations,
+        filter_func=filter_func,
+        compare_distinct=compare_distinct,
+        output_processors=output_processors,
+        condition=condition,
+        map_func=function,
+        # cache_size=0,
```
0xbe7a, line 1156:
```suggestion
```

tests/integration/test_integration.py
0xbe7a: Can't you also use the same query collector to check that nothing has been cached?

SimonLangerQC: Done.

SimonLangerQC: yes, I can additionally add it as well :) - Done

src/datajudge/constraints/row.py
```diff
@@ -34,7 +35,7 @@ def test(self, engine: sa.engine.Engine) -> TestResult:
 
 
 class RowEquality(Row):
-    @lru_cache(maxsize=None)
+    # @lru_cache(maxsize=None)
```
0xbe7a, line 38:
```suggestion
```

src/datajudge/constraints/row.py
```diff
@@ -45,7 +46,7 @@ def get_factual_value(self, engine: sa.engine.Engine) -> Tuple[int, int]:
         self.factual_selections = [*selections_left, *selections_right]
         return n_rows_missing_left, n_rows_missing_right
 
-    @lru_cache(maxsize=None)
+    # @lru_cache(maxsize=None)
```
0xbe7a, line 48:
```suggestion
```

tests/integration/test_integration.py
```diff
@@ -1,8 +1,11 @@
+from __future__ import annotations
```
0xbe7a, line 1: Do we need this?

tests/unit/test_utils.py
0xbe7a: It seems like you want to make sure the `output_processor_limit` does not mutate its arguments, right? I think there should be a comment explaining this

0xbe7a: Also, off-topic and not strictly about this PR: I think the naming is pretty confusing. I had to read the PR for the OutputProcessor to figure out a) the problem it is solving and b) what it’s doing.

Questions like:
"Is this outputting the processor limit?"
"Is this a processor, as in CPU?"

I think OutputFormatter would be a better name, and maybe `truncate_output` would be a better name for `output_processor_limit`.

tests/integration/test_integration.py
SimonLangerQC: Yes, since `QueryCollector` references itself

tests/unit/test_utils.py
SimonLangerQC: We originally also had a different name for it, and considered saying postprocessor instead of just processor. These changes are only visible in this PR due to the afore-mentioned merge conflicts; changing their names now would be breaking changes from the previous 1.9.1 release.

https://github.com/Quantco/datajudge/pull/224#discussion_r1627624676

SimonLangerQC: Added an explaining comment.

SimonLangerQC: The user-facing documentation is available in Uniques.py: https://github.com/Quantco/datajudge/blob/a935d64bc299b6b9096024fcf6ab87e0fefcd1b6/src/datajudge/constraints/uniques.py#L91

0xbe7a (approved):

PR #241, opened 2024-09-03
Fix a bug with the current implementation of the `NumericPercentile` `Constraint`

kklein:
Previously, the percentile computation was not according to the definition we provided ourselves. 

Our definition said that
> The percentile is defined as the smallest value present in ``column`` for which `percentage` % of the values in `column` are less or equal.

Let's now assume we have the following data

| value | rank | rank/n |
|---|---|---|
| 1 | 1 | 1/19 ~ .05 |
| 2 | 2 | 2/19 ~ .11 |
| 3 | 3 | 3/19 ~ .16 |
| 4 | 4 | 4/19 ~ .21 |
| ... | ... | ... | 
| 19 | 19 | 19/19 = 1 |

If we set the percentile of interest to be 20%, we see that the value 4 is the smallest value present in the data, for which at least 20% of the data is lesser or equal to that value.

Previously, our retrieval would've yielded 3. This behavior is explicitly reflected in our own tests.


ivergara (approved):
If you plan on releasing right away you can fix the changelog file before merging.

PR #255, opened 2024-12-23
Address mssql CI failure

kklein:
# Status quo

- Our mssql integration tests currentlhy support two drivers.
- Our CI integration tests only run tests against one of both.
- CI on main is failing for mssql

# Changes

This PR makes CI use the the other type of driver. While this doesn't solve the problem with the previous driver, it makes for a simpler set up and makes tests run through.

I'm very open to suggestions to go about this in a different way!


ivergara (approved):
Labeled ready but the PR is still in draft.

PR #256, opened 2025-01-12
Bugfix: `NoOverlap2d` constraint fails to recognize duplicates

kklein:
Bug detected by @EgeKaraismailogluQC.

Failing example: https://github.com/Quantco/datajudge/pull/256/commits/79aff3bfae0c15a9a56da11486834201c717e9a7

Fundamentally, the overlap detection contained a flaw in which the 'left' bound of two intervals could not be equal for them to be considered overlapping. A special case of this situation is when two intervals are exactly equal - the counterexample which made this bug apparent.

Importantly, this bug affects all overlap-related constraints.

The bug is fixed in `datajudge.db_access.get_interval_overlaps_nd` by:
- Replacing a strict inequality by an equality
- Explicitly handling the case where an n-dimensional interval is duplicated; if we didn't handle this explicitly we would always find an 'overlap' for a table is joined with itself

Test cases are added for all relevant overlap-constraint-adding methods.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,8 @@ Changelog
 **Bug fixes**
 
 - Fix a bug in :meth:`datajudge.WithinRequirement.add_date_no_overlap_constraint`
-  and :meth:`datajudge.WithinRequirement.add_date_no_overlap2d_constraint` in which
+  and :meth:`datajudge.WithinRequirement.add_date_no_overlap2d_constraint` and
+  :meth:`datajudge.WithinRequirement.add_numeric_no_overlap_constraint` in which
   some overlaps were not detected due to equality of their leftmost bounds.
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,16 @@
 Changelog
 =========
 
+1.9.3 - 2025.01.13
+------------------
+
+**Bug fixes**
+
+- Fix a bug in :meth:`datajudge.WithinRequirement.add_date_no_overlap_constraint`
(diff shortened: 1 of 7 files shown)


kklein:
@ivergara I believe it is necessary to manually tear down some integration test data, since the content of some tables has changed and we don't have an automatic teardown for snowflake. Would you be so kind to do this @ivergara ? I don't think I have access to (and if so I have forgotten how to) enter the snowflake administration interface.

EgeKaraismailogluQC:
@kklein thanks for tagging me here! I am not familiar enough with the codebase to approve or request changes, but I do have one question: why do we need to handle an exact match separately? It seems to me that our definition of an overlap in `naive_violation_condition` should cover exact matches.

kklein:
Thanks for taking a look @EgeKaraismailogluQC .

> why do we need to handle an exact match separately? 

In order to determine whether there are overlaps within a table, we merge the table onto itself with the 'overlap condition' as a join condition. Hence `naive_violation_condition` will always find a violation: a row overlaps with itself.

Does that make sense?

ivergara:
> @ivergara I believe it is necessary to manually tear down some integration test data, since the content of some tables has changed and we don't have an automatic teardown for snowflake. Would you be so kind to do this @ivergara ? I don't think I have access to (and if so I have forgotten how to) enter the snowflake administration interface.

Will take a look. I do still know how to log in, and I hope that by deleting the data it'll be enough.

EgeKaraismailogluQC:
@kklein makes sense, thanks!

kklein:
Thanks @ivergara - the tests are passing now :)
Do you think you could review the PR?

ivergara (approved):

PR #262, opened 2025-02-04
Add type hints to `db_access`

kklein:


src/datajudge/constraints/row.py
```diff
@@ -36,6 +38,8 @@ def test(self, engine: sa.engine.Engine) -> TestResult:
 
 class RowEquality(Row):
     def get_factual_value(self, engine: sa.engine.Engine) -> Tuple[int, int]:
+        if self.ref is None or self.ref2 is None:
```
kklein, line 41: This is a bit annoying. The `Row` constraints are always between several data sources and thereby `ref` and `ref2` should never be None. One can easily see in their initialization function, that the inputs to the initialization have according type hints. Yet, mypy doesn't seem to pick this up.

src/datajudge/db_access.py
```diff
@@ -351,37 +360,39 @@ def get_column(self, engine):
...
             return lowercase_column_names(self.columns)
         return self.columns
 
-    def get_columns_or_pk_columns(self, engine):
+    def get_columns_or_pk_columns(self, engine: sa.engine.Engine) -> list[str] | None:
         return (
             self.columns
             if self.columns is not None
-            else get_primary_keys(engine, self.data_source)
+            else get_primary_keys(engine, self)[0]
```
kklein, line 375: This was actually a bug we found through mypy :)

src/datajudge/constraints/row.py
ivergara: Don't know a better way.

On another note, the line before has a Tuple return type instead of tuple.

ivergara (approved):

src/datajudge/constraints/row.py
kklein: Great catch!

PR #263, opened 2025-02-05
Consistently use Python 3.10 type annotations

kklein:

### DIFF
--- a/src/datajudge/constraints/base.py
+++ b/src/datajudge/constraints/base.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import abc
 from dataclasses import dataclass, field
 from functools import lru_cache
-from typing import Any, Callable, Collection, List, Optional, Tuple, TypeVar, Union
+from typing import Any, Callable, Collection, List, Optional, TypeVar
 
 import sqlalchemy as sa
 
@@ -16,7 +18,7 @@ OptionalSelections = Optional[List[sa.sql.expression.Select]]
 ToleranceGetter = Callable[[sa.engine.Engine], float]
 
 
-def uncommon_substrings(string1: str, string2: str) -> Tuple[str, str]:
+def uncommon_substrings(string1: str, string2: str) -> tuple[str, str]:
     qualifiers1 = string1.split(".")
     qualifiers2 = string2.split(".")
     if qualifiers1[0] != qualifiers2[0]:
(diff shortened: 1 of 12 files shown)


src/datajudge/constraints/base.py
```diff
@@ -121,12 +123,12 @@ def __init__(
...
-        name: Optional[str] = None,
-        output_processors: Optional[
-            Union[OutputProcessor, List[OutputProcessor]]
-        ] = output_processor_limit,
+        ref2: DataReference | None = None,
+        ref_value: Any = None,
+        name: str | None = None,
+        output_processors: OutputProcessor
+        | list[OutputProcessor]
+        | None = output_processor_limit,
```
ivergara, line 131: Why a None did sneak in here?

src/datajudge/constraints/uniques.py
```diff
@@ -106,21 +108,21 @@ class Uniques(Constraint, abc.ABC):
...
         ref: DataReference,
-        name: Optional[str] = None,
+        name: str | None = None,
         cache_size=None,
-        output_processors: Optional[
-            Union[OutputProcessor, List[OutputProcessor]]
-        ] = output_processor_limit,
+        output_processors: OutputProcessor
+        | list[OutputProcessor]
+        | None = output_processor_limit,
```
ivergara, line 115: Same here

src/datajudge/constraints/base.py
```diff
@@ -121,12 +123,12 @@ def __init__(
...
         ref: DataReference,
         *,
-        ref2: Optional[DataReference] = None,
-        ref_value: Optional[Any] = None,
-        name: Optional[str] = None,
-        output_processors: Optional[
-            Union[OutputProcessor, List[OutputProcessor]]
-        ] = output_processor_limit,
+        ref2: DataReference | None = None,
+        ref_value: Any = None,
```
ivergara, line 127: Not critical, but the change from optional any to any feels wrong. But probably is an issue that Any is too vague to start with.

ivergara (approved):

src/datajudge/constraints/base.py
kklein: I see your point. Do you suggestion yet?

src/datajudge/constraints/base.py
kklein: Before this change, `output_processors` annotated with `Optional[Union[OutputProcessor, List[OutputProcessor]]`. 

If we convert the `Optional`, this given us:
`Union[OutputProcessor, List[OutputProcessor] | None`.

If we now also covert the `Union` and (`List` to `list`), this gives us `OutputProcessor | list[OutputProcessor] | None`, which, afaict corresponds to the suggested change.

Does that make sense?

src/datajudge/constraints/base.py
ivergara: I have no suggestion. Well, one, but it's quite vague. To create a type that reflects what a `ref_value` can be. But it feels too much work for little gain.

src/datajudge/constraints/base.py
ivergara: Seems I missed a step in the chain of transformations.

PR #269, opened 2025-03-11
Remove bound on ibm_db.

kklein:
We introduced a bound on `ibm_db` in https://github.com/Quantco/datajudge/pull/260 due to issues with recent versions of the package.

These issues have faded since https://github.com/conda-forge/ibm_db-feedstock/pull/81.

Therefore this PR removes the upper bound on `ibm_db` again.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -54,8 +54,7 @@ test = "pytest tests/integration --backend=postgres"
 coverage = "pytest tests/integration --cov=datajudge --cov-report=xml --cov-append --backend=postgres"
 
 [feature.db2.dependencies]
-# Upper bounded due to https://github.com/ibmdb/python-ibmdb/issues/986
-ibm_db = "<=3.2.4"
+ibm_db = "*"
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -54,8 +54,7 @@ test = "pytest tests/integration --backend=postgres"
 coverage = "pytest tests/integration --cov=datajudge --cov-report=xml --cov-append --backend=postgres"
 
 [feature.db2.dependencies]
-# Upper bounded due to https://github.com/ibmdb/python-ibmdb/issues/986
-ibm_db = "<=3.2.4"
+ibm_db = "*"
```
ivergara, line 57: Shouldn't it be better to scope the misbehaving version(s) out?

kklein: You're right of course!
[9022938](https://github.com/Quantco/datajudge/pull/269/commits/90229385ee7d68bab3d6e3772a119ebbcd27cdb9)

ivergara (approved):

